PR #9, opened 2026-03-02
docs: add README

pdscomp:
Refs #1


jamesturton:
Also we can update the readme to use `sunxi-fel uboot u-boot-sunxi-with-spl.bin` as the boot command rather than the 3 allwinner-xfel commands. `sunxi-tools` should also be added to the prerequisites as this provides `sunxi-fel`

PR #20, opened 2026-03-15
lcd: disable screen in addition to backlight to prevent ghosting

timiv:
The LCD panel suffered from image retention (ghosting) after the screen went dark during guppyscreen's sleep mode.

Two problems combined to cause this:

1. The DTS had enable-gpios on the panel node pointing to PD22 (the backlight pin), while PB8 (panel enable) was never touched. As a result, the LCD module stayed permanently powered.
2. guppyscreen called FB_BLANK_NORMAL, which only turns off the backlight. The DRM DPMS Off path, which also pulls the panel power pin LOW and disables the TCON, was never triggered.
With the backlight off but the panel still powered and receiving no refresh signals, the LC cells retained charge from the last displayed image.

### DIFF
--- a/meta-opencentauri/recipes-kernel/linux/linux-mainline/elegoo-centauri-carbon1/elegoo-centauri-carbon1.dts
+++ b/meta-opencentauri/recipes-kernel/linux/linux-mainline/elegoo-centauri-carbon1/elegoo-centauri-carbon1.dts
@@ -48,21 +48,18 @@
 		vin-supply = <&reg_vcc5v>;
 	};
 
-	// backlight: backlight {
-	// 	compatible = "pwm-backlight";
-	// 	pwms = <&pwm 7 20000 0>;
-	// 	brightness-levels = <0 4 8 16 32 64 128 255>;
-	// 	default-brightness-level = <7>;
-	// 	enable-gpios = <&pio 1 8 GPIO_ACTIVE_HIGH>; /* PB8 */
-	// };
+	backlight: backlight {
+		compatible = "gpio-backlight";
+		gpios = <&pio 3 22 GPIO_ACTIVE_HIGH>; /* PD22 */
+		default-on;
+	};
 
 	panel {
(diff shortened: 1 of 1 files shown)


meta-opencentauri/recipes-apps/guppyscreen/files/0001-lv_driver_fb_ioctls.patch
```diff
@@ -7,7 +7,7 @@ index fe8d2ff..e8f0928 100644
  }
  
 +void fbdev_blank(void) {
-+  if (ioctl(fbfd, FBIOBLANK, FB_BLANK_NORMAL) == -1) {
++  if (ioctl(fbfd, FBIOBLANK, FB_BLANK_POWERDOWN) == -1) {
```
jamesturton: This patch file (and some of the others in this directory) were actually taken from the guppyscreen source https://github.com/ballaswag/guppyscreen/blob/main/patches/0001-lv_driver_fb_ioctls.patch
I rather lazily copied them in so that yocto would patch the source automatically without having to append to the do_patch task... I can't see it being a problem, but we just need to remember that it's not the stock patch in case it gets changed in the future.

timiv: Ah, didn’t realize that. 

I did check the register contents directly after the ioctl and saw the relevant change not being changed with the original ioctl. 

But maybe I overlooked something and we can just leave it as it is.

Let’s just keep the DTS in this PR. I’ll update it.  If necessary I’ll provide a clean patch for the ioctl in a separate PR.

jamesturton: Sure! Up to you, I'm happy to merge as is if you don't want to go to the effort of splitting the PR? :sweat_smile:

PR #22, opened 2026-03-17
Wait 3s after toolhead bootloader flash

suchmememanyskill:
Maybe we should just check if /dev/ttyACM0 exists for faster exec time, but this will probably work too.

Untested.


jamesturton:
3 seconds seems to be a good amount of time to wait.
I've re-jigged the init script to speed-up our boot time and included this wait after bootloader flashing too :smile: 
a1509298e4141259085eb8df9aef55c50323d615

PR #27, opened 2026-03-23
Add update scripts

jamesturton:


meta-opencentauri/recipes-data/update-scripts/files/switch-to-stock
```diff
@@ -0,0 +1,37 @@
+#!/bin/sh
+
+set -e
+
+# ver 1.1.46
```
suchmememanyskill, line 5: I would switch to patched oc instead of stock.

If anything blows up in the upgrade process, on stock recovery is insanely tricky. Under patched OC we have options.

meta-opencentauri/recipes-data/update-scripts/files/switch-to-stock
```diff
@@ -0,0 +1,37 @@
...
+
+# Decrypt firmware
+exec "$DECRYPT" "$BINFILE" "$KEY" "$IV" "$SWUFILE"
+
+# Install firmware
+flash "$SWUFILE"
+
+# Flash stock bootloader
+service klipper stop
+flashtool -d $SERIALPORT_BED -b 115200 -r
```
suchmememanyskill, line 31: I think it's a lot better to do it here as well. Patching this in in the patched firmware is quite a mess

meta-opencentauri/recipes-data/update-scripts/files/update-cosmos
```diff
@@ -0,0 +1,17 @@
+#!/bin/sh
```
suchmememanyskill, line 1: Question, does guppy/grumpy show stdout/stderr while its running the script?

meta-opencentauri/recipes-data/update-scripts/files/factory-reset
```diff
@@ -0,0 +1,6 @@
+#!/bin/sh
+
+sleep 3
+touch /data/.factory-reset
```
suchmememanyskill, line 4: How does this work? I assume another script somewhere else not in this PR?

meta-opencentauri/recipes-data/update-scripts/files/switch-to-stock
```diff
@@ -0,0 +1,37 @@
+#!/bin/sh
+
+set -e
+
+# ver 1.1.46
+FW_URL="https://download.chitubox.com/chitusystems/chitusystems/public/printer/firmware/release/1/ca8e1d9a20974a5896f8f744e780a8a7/1/01.01.46/2025-10-22/f9bd2b9b1926408ca238de8e7eac69b6.bin"
+DECRYPT="/usr/bin/swu-decrypt.py"
+KEY="78B6A614B6B6E361DC84D705B7FDDA33C967DDF2970A689F8156F78EFE0B1FCE"
```
suchmememanyskill, line 7: Hm, unsure if plaintext key here is the best approach. Although don't think elegoo particuarly cares (or anycubic for that matter)

meta-opencentauri/recipes-data/update-scripts/files/switch-to-stock
jamesturton: We could implement both and let the user choose?
Then it's only the case of changing the URL to download from and skip decrypting it.

meta-opencentauri/recipes-data/update-scripts/files/update-cosmos
jamesturton: Nope, it just looks at the exit code, however it shouldn't return if it successful as it should reboot.
Not as nice as your recovery app :wink:

meta-opencentauri/recipes-data/update-scripts/files/factory-reset
jamesturton: It's done in the custom init script here
https://github.com/OpenCentauri/yocto-opencentauri/blob/4e1243b0da772f98a05a364498afa8ee51e761c8/meta-opencentauri/images/files/overlayfs-etc-preinit.sh.in#L37

meta-opencentauri/recipes-data/update-scripts/files/switch-to-stock
jamesturton: Well... it already exists on the internet...

PR #33, opened 2026-03-26
config: Improve homing

timiv:
The current homing macros had a few flaws:

1. If a bed mesh was loaded klipper aborted homing of Z if the G1 move was not yet complete. This is fixed by adding an additional M400.

2. It was not possible to home X / Y individually as the second home move was always done.

3. Homing to the front right side requires Y to be homed before X as otherwise the toolhead might crash into the lever of the poop tray, if it starts there.

### DIFF
--- a/meta-opencentauri/recipes-apps/klipper/files/mainboard.cfg
+++ b/meta-opencentauri/recipes-apps/klipper/files/mainboard.cfg
@@ -18,7 +18,7 @@ position_endstop: 256.499
 position_max: 256.5
 position_min: -2.01
 homing_speed: 80
-homing_retract_dist: 3
+homing_retract_dist: 20
 rotation_distance: 39.872
 
 [stepper_y]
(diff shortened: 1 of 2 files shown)


meta-opencentauri/recipes-apps/klipper/files/mainboard.cfg
```diff
@@ -18,7 +18,7 @@ position_endstop: 256.499
 position_max: 256.5
 position_min: -2.01
 homing_speed: 80
-homing_retract_dist: 3
+homing_retract_dist: 20
```
Atomique13, line 21: The homing_retract_dist setting must be set to zero in the stepper_x config section to disable the second homing move. The second homing attempt does not add value when using sensorless homing, it will not work reliably, and it will confuse the tuning process.

per klipper docs and also tested by me.

meta-opencentauri/recipes-apps/klipper/files/mainboard.cfg
```diff
@@ -32,7 +32,7 @@ position_endstop: -2.5
 position_max: 270.000000
 position_min: -2.51
 homing_speed: 80
-homing_retract_dist: 3
+homing_retract_dist: 30
```
Atomique13, line 35: same as early

PR #38, opened 2026-03-28
Load/Unload macro

suchmememanyskill:


meta-opencentauri/recipes-apps/klipper/files/macros.cfg
```diff
@@ -255,14 +287,12 @@ gcode:
     G0 X255 F8000                  ; Move to cutting position
     G0 Y3 F1200                     ; Cut
```
Atomique13, line 288: you can retract a bit of filament before cutting to save some, especially when multi color if it uses same macro

PR #49, opened 2026-04-01
Update URL for OC v0.3.0 update.sw

jaimbo:
Fix the FW_URL for switching back from COSMOS to OC

### DIFF
--- a/meta-opencentauri/recipes-data/update-scripts/files/switch-to-oc-patched
+++ b/meta-opencentauri/recipes-data/update-scripts/files/switch-to-oc-patched
@@ -14,6 +14,13 @@ STOCK_FW_TOOLHEAD="/lib/firmware/katapult-deployer-stock-toolhead.bin"
 # Download firmware
 curl -k -f -S -o "$SWUFILE" -L "$FW_URL"
 
+# Check file exists and is not empty, clean up and exit if not
+if [ ! -s "$SWUFILE" ]; then
+    echo "Failed to download update.swu - Aborting."
+    rm -f "$SWUFILE"
+    exit 1
+fi
+
 # Install firmware
 flash "$SWUFILE"
 


meta-opencentauri/recipes-data/update-scripts/files/switch-to-oc-patched
```diff
@@ -2,7 +2,7 @@
 
 set -e
 
-FW_URL="https://github.com/OpenCentauri/yocto-opencentauri/releases/latest/download/update.swu"
+FW_URL="https://github.com/OpenCentauri/cc-fw-tools/releases/download/v0.3.0/update.swu"
```
jamesturton, line 5: Good catch! 😄 
Is there any reason you chose to switch to a fixed release version rather than using latest? It's just something else we will have to remember to update on the next oc-patched release if we use a fixed version.

meta-opencentauri/recipes-data/update-scripts/files/switch-to-oc-patched
```diff
@@ -12,7 +12,14 @@ SERIALPORT_TOOLHEAD="/dev/serial/by-path/platform-4101400.usb-usb-0:1:1.0"
 STOCK_FW_TOOLHEAD="/lib/firmware/katapult-deployer-stock-toolhead.bin"
 
 # Download firmware
-curl -k -o "$SWUFILE" -L "$FW_URL"
+curl -k -f -S -o "$SWUFILE" -L "$FW_URL"
+
+# Check file exists and is not empty, clean up and exit if not
+if [ ! -s "$SWUFILE" ]; then
```
jamesturton, line 18: I guess this is a little redundant as if there is no file existing there will be nothing to remove and if it's empty then swupdate will fail very early. Even if the .swu file doesn't download completely swupdate will catch this when it checks the md5 sum of the payload against the header. Or have I missed something?

jamesturton: Curl failing on 4xx errors is nice to have though!

meta-opencentauri/recipes-data/update-scripts/files/switch-to-oc-patched
jaimbo: No reason beyond wanting to get a quick fix in place and not fully considering the best options. I think that would be a very sensible change!

PR #52, opened 2026-04-01
Fix moonraker warnings

### LINKED ISSUE #44: Several Moonraker warnings upon bootup
```
Component 'machine' failed to load with error: The name org.freedesktop.systemd1 was not provided by any .service files
```
```
Supplied path (/usr/share/klipper/config) for (config_examples) is invalid. Make sure
that the path exists and is not the file system root.
```
```
Supplied path (/usr/share/klipper/docs) for (docs) is invalid. Make sure
that the path exists and is not the file system root.

```
Moonraker: machine
An error was detected while loading the moonraker component 'machine'. Please check the log file and fix the issue.
```

jamesturton:
Tested working on my machine. Whether we need to actually include the docs and configs is dubious, but for now it should be ok till we start running out of space on the squashfs

Fixes #44


suchmememanyskill:
I'd argue why include the docs/configs at that point, if it may cause a headache later. But pr itself looks fine

PR #55, opened 2026-04-01
optimized python memory usage by disabling unused modules

niklasarnitz:
Proof of concept to reduce the memory footprint of python.
I can't test it on my printer at the moment since I'm on vacation now.
But it builds.


jamesturton:
Well it doesn't seem to cause any regressions which is great!
I saw a 2~3% reduction in ram during idle when running this change, but I'm honestly not sure how much variation there is between boots so that might just be within the normal variation.

As I'm not a super python geek, is there a way that I can check that python is actually using the pre-compiled .pyc bitcode? I guess I could try removing all the .py files just to test, but you said that there was still some dependency on the .py file iirc?

PR #62, opened 2026-04-02
Stabilize Moonraker/Klippy: zram default + BusyBox-compatible ip JSON shim

shawn-makes-stuff:
## Summary
This PR adds two stability fixes observed to reduce Moonraker/Klippy dropouts on Centauri Carbon Cosmos 0.0.2:

- Add a BusyBox-compatible `ip` shim at `${datadir}/moonraker/ip` that supports `ip -json -det address`
- Ensure Moonraker uses that shim by prepending `/usr/share/moonraker` in `moonraker-init-d` `PATH`
- Set zram default to `FACTOR=400` directly in the zram init script (single default source)

## Why
On stock image behavior, Moonraker network parsing calls `ip -json -det address`, but BusyBox `ip` does not support those flags. This repeatedly raises shell command errors and contributes to instability under load.

At the same time, memory pressure can kill or destabilize Python services. Raising zram factor provides more compressed swap capacity and reduces OOM risk.

## Files changed
- `meta-opencentauri/recipes-apps/moonraker/files/moonraker-init-d`
- `meta-opencentauri/recipes-apps/moonraker/files/ip` (new)
- `meta-opencentauri/recipes-apps/moonraker/moonraker_0.10.0.bb`
- `meta-opencentauri/recipes-core/zram/files/init`
- `meta-opencentauri/recipes-core/zram/files/zram-default` (removed)
- `meta-opencentauri/recipes-core/zram/zram_%.bbappend`

## Validation
- `server/info` reports `klippy_connected=true` and `klippy_state=ready`
- Moonraker `ip -json -det address` path succeeds via shim
- zram swap active after service start with expected capacity increase


meta-opencentauri/recipes-core/zram/zram_%.bbappend
```diff
@@ -1 +1,7 @@
 FILESEXTRAPATHS:prepend := "${THISDIR}/files:"
+SRC_URI += "file://zram-default"
+
+do_install:append() {
+    install -d ${D}${sysconfdir}/default
+    install -m 0644 ${WORKDIR}/zram-default ${D}${sysconfdir}/default/zram
```
jamesturton: I think it's best to change the default in the init script
https://github.com/OpenCentauri/yocto-opencentauri/blob/main/meta-opencentauri/recipes-core/zram/files/init#L15
Having two default values will get a little confusing

meta-opencentauri/recipes-apps/moonraker/moonraker_0.10.0.bb
```diff
@@ -78,11 +79,16 @@ do_install() {
     # Install SysVinit script
     install -d ${D}${sysconfdir}/init.d
     install -m 0755 ${WORKDIR}/moonraker-init-d ${D}${sysconfdir}/init.d/moonraker
+
+    # Install BusyBox-compatible ip shim used by Moonraker network probing
+    install -d ${D}${sysconfdir}/klipper/bin
+    install -m 0755 ${WORKDIR}/ip ${D}${sysconfdir}/klipper/bin/ip
```
jamesturton: I'm not sure `sysconfdir` is the right place for this as we shouldn't need it to be configurable in the overlayfs. Maybe it can go in `${datadir}/moonraker/` alongside the other moonraker python scripts?

shawn-makes-stuff: Addressed in 6246807: moved shim install to ${datadir}/moonraker/ip and updated moonraker-init-d PATH to prepend /usr/share/moonraker.

meta-opencentauri/recipes-core/zram/zram_%.bbappend
shawn-makes-stuff: Addressed in 6246807: removed zram-default and set FACTOR=400 directly in recipes-core/zram/files/init so there is only one default source.

shawn-makes-stuff:
Note: this is ✨vibe code✨ - I do my best to make sure its logical and functional, tested and working on my machine, but do verify for yourself that nothing odd pops up.

PR #67, opened 2026-04-02
Reintroduce doom

suchmememanyskill:
Tested, fully works both via touch and via keyboard


meta-opencentauri/recipes-apps/klipper/files/gcode_shell_command.py
```diff
@@ -0,0 +1,95 @@
+# Run a shell command via gcode
```
suchmememanyskill: This is unnecessary, we already include this in our kalico fork

meta-opencentauri/recipes-kernel/linux/linux-mainline/elegoo-centauri-carbon1/kernel-size-reduction.cfg
```diff
@@ -96,8 +96,8 @@ CONFIG_RFKILL=y
 # ============================================================
 # No keyboard/mouse input (keep touchscreen via fragment.cfg)
 # ============================================================
-# CONFIG_INPUT_KEYBOARD is not set
-# CONFIG_INPUT_MOUSE is not set
+CONFIG_INPUT_KEYBOARD=y
+CONFIG_INPUT_MOUSE=y
```
suchmememanyskill, line 100: unsure if this does anything.

PR #70, opened 2026-04-03
klipper: Improve performance

timiv:
This slightly improves the performance of klippy by disabling assertions and debug statements in python.


timiv:
After reviewing #55, I decided to add `-OO` . We either need to integrate this with #55 or merge it before. #55 only compiles python modules with `-OO` but does not seem to call python with proper flags.

PR #72, opened 2026-04-03
klipper: add default filament runout sensor config

shawn-makes-stuff:
At a basic level, it does this:

  - enables Klipper’s filament_switch_sensor
  - maps it to the printer’s runout signal on PC0
  - makes the printer report filament present / absent through the normal Klipper sensor path

Make the stock filament sensor available in the config so the printer can detect filament status and use normal runout behavior. This is config only, it does not enable filament run out logic.

### DIFF
--- a/meta-opencentauri/recipes-apps/klipper/files/mainboard.cfg
+++ b/meta-opencentauri/recipes-apps/klipper/files/mainboard.cfg
@@ -122,42 +122,6 @@ idle_speed: 0
 [filament_switch_sensor filament_sensor]
 # Centauri Carbon stock runout sensor is wired to the mainboard, so keep the
 # hardware definition in klipper-readonly alongside the other board config.
+# leave pause policy to the normal Klipper sensor settings.
 switch_pin: PC0
 pause_on_runout: False
-
-[gcode_macro _FILAMENT_SENSOR_EVENT]
-description: Hook for filament sensor transitionr
-gcode:
-    {% set state = params.STATE|default("")|lower %}
-    {% if state == "inserted" %}
-        # Log to the console instead of the LCD so the message does not linger.
-        {action_respond_info("Filament inserted")}
-    {% elif state == "removed" %}
-        # Keep this as a notification hook only. Any pause/runout policy should be opt-in.
-        {action_respond_info("Filament removed")}
-    {% endif %}
-
-[gcode_macro _FILAMENT_SENSOR_WATCH]
-# Poll the hardware sensor and turn state changes into a single hook call.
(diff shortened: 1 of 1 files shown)


meta-opencentauri/recipes-apps/klipper/files/printer.cfg
```diff
@@ -1,4 +1,5 @@
 [include klipper-readonly/*.cfg]
+[include filament-sensor.cfg]
```
suchmememanyskill: For future reference, everything we configure should be in klipper-readonly

meta-opencentauri/recipes-apps/klipper/files/filament-sensor.cfg
```diff
@@ -0,0 +1,9 @@
+# External filament runout sensor (Centauri Carbon stock path)
+[filament_switch_sensor filament_sensor]
```
suchmememanyskill: I think we should put this in mainboard.cfg, as it's connected directly to the mainboard

meta-opencentauri/recipes-apps/klipper/files/filament-sensor.cfg
```diff
@@ -0,0 +1,9 @@
...
+# Important: Centauri Carbon stock signal is PC0 (raw, no pull-up ^) to avoid false always-detected state.
+switch_pin: PC0
+pause_on_runout: False
+runout_gcode:
+  M117 Filament removed
```
suchmememanyskill: Could SET_DISPLAY_TEXT be used instead?

Although, unsure to be honest if it should be such a status message. Maybe a respond action is good enough so it gets logged to the console.

suchmememanyskill: {action_respond_info("message")}

meta-opencentauri/recipes-apps/klipper/files/printer.cfg
shawn-makes-stuff: I’ve moved the sensor config into klipper-readonly/mainboard.cfg and I’ve kept all configuration there rather than in a separate file

meta-opencentauri/recipes-apps/klipper/files/filament-sensor.cfg
shawn-makes-stuff: I also switched the user-facing notifications to action_respond_info() instead of SET_DISPLAY_TEXT, since this is better as a console/status event and avoids leaving sticky text on the LCD

PR #79, opened 2026-04-04
atomscreen: ensure target sysroot for linuxfb

colin-campbell:
On Ubuntu 24.04 at least, generating the FFI bindings for linuxfb in the atomscreen recipe fails because bindgen uses the host sysroot instead of the target sysroot.

This PR adds explicit bindgen args to the atomscreen recipe to ensure that the target sysroot is used.


jamesturton (changes requested):
Using export in the recipe here makes the variable available to the entire shell environment for all tasks, which is too broad. The correct approach is to set it as a tasklocal variable preferably scoped only to the compile task.

colin-campbell:
In that case it shouldn't go in this repo at all, I suppose? 
It'll have to in a .env passthrough in Cargo.toml for atomscreen/driver which has linuxfb as a dependency, right?  

I'll close this.

jamesturton:
Someone else also had this problem #253 so I figured it was time to revisit this :)
I was wrong about how I though the export is used, apparently bitbake runs as its own separate spawned shell process for each recipe so doing this is safe. I found how an other meta-openembedded recipe handles this error and I tweaked your changes to also include the target architecture, however I don't think it matters so much in this case.
https://patchwork.yoctoproject.org/project/oe/patch/20250227183238.1909597-1-skandigraun@gmail.com/
I cherry-picked and merged your commit here https://github.com/OpenCentauri/cosmos/commit/75b2c1744dcfe39d74b0330ee27e090fe6888194, thanks for the contribution!

PR #82, opened 2026-04-04
Nozzle Wipe to Resume Macro

gta4hater:
It cleans the nozzle when the print resumes.

### DIFF
--- a/meta-opencentauri/recipes-apps/klipper/files/macros.cfg
+++ b/meta-opencentauri/recipes-apps/klipper/files/macros.cfg
@@ -452,6 +452,17 @@ gcode:
     {% set bed_temp = params.BED_TEMP|default(60)|float %}
     BED_MESH_CLEAR
     M104 S140
+
+[gcode_macro RESUME]
+rename_existing: BASE_RESUME
+gcode:
+    {% set EXTRUDER_TEMP = printer.extruder.target|int %}
+    M106 P3 S255
+    M109 S200
+    M729
+    M109 S{EXTRUDER_TEMP}
+    M106 P3 S0
+    BASE_RESUME
     M400
     G28
     M109 S140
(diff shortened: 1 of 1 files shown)


meta-opencentauri/recipes-apps/klipper/files/macros.cfg
```diff
@@ -457,3 +457,14 @@ gcode:
     M109 S140
     M729
     SCREWS_TILT_CALCULATE
+
+[gcode_macro RESUME]
+rename_existing: BASE_RESUME
+gcode:
+    {% set EXTRUDER_TEMP = printer.extruder.target|int %}
```
suchmememanyskill: When do you set this? I assume this won't work when you press resume in the webui or grumpy

meta-opencentauri/recipes-apps/klipper/files/macros.cfg
```diff
@@ -457,3 +457,14 @@ gcode:
     M109 S140
     M729
     SCREWS_TILT_CALCULATE
+
+[gcode_macro RESUME]
```
suchmememanyskill, line 476: Wouldn't it be better to write this in the mainsail/fluidd cfg's instead of overriding resume?

meta-opencentauri/recipes-apps/klipper/files/macros.cfg
gta4hater: i mean i only tested it in the printer.cfg

meta-opencentauri/recipes-apps/klipper/files/macros.cfg
gta4hater: didnt think of that...

meta-opencentauri/recipes-apps/klipper/files/macros.cfg
gta4hater: via mainsail it worked but i cant tell you if it would work in the macro.cfg

gta4hater: via guppy it also works tested it

suchmememanyskill:
No longer relevant due to macros PR

PR #88, opened 2026-04-05
Add thermal zone configuration for CPU monitoring

Atomique13:
## Summary

Adds full CPU thermal monitoring for the Elegoo Centauri Carbon 1 (R528/T113-S3 SoC) using the on-chip Thermal Hardware Sensor (THS) block.

### Device Tree (`elegoo-centauri-carbon1.dts`)
- Adds `ths` thermal sensor node at `0x02009400` with compatible `allwinner,sun20i-d1-ths`
- Adds `cpu-thermal` thermal zone bound to the THS sensor, with two trip points:
  - **85°C passive** — thermal alert threshold (monitoring only; no cooling device bound — see rationale below)
  - **100°C critical** — emergency shutdown

### Kernel Backport Patch (`0003-thermal-sun8i-add-sun20i-d1-ths-support.patch`)
Backports `sun20i-d1-ths` support to the `sun8i_thermal` driver in Linux 6.6.85. The R528/T113-S3 uses the same THS IP block as the Allwinner D1/T113 line (`sun20i-d1-ths`), which is natively supported upstream in later kernels. The patch adds:
- `sun20i_d1_ths` chip description struct (1 sensor, `sun50i_h6`-compatible calibration/init/irq)
- OF match entry for `allwinner,sun20i-d1-ths`

### Kernel Config (`fragment.cfg`)
Adds explicit thermal framework configuration alongside the existing `CONFIG_SUN8I_THERMAL=y` driver config:

| Config | Purpose |
|--------|---------|
| `CONFIG_THERMAL=y` | Base thermal framework |
| `CONFIG_THERMAL_OF=y` | Device-tree thermal zone parsing (required for DTS nodes to work) |
| `CONFIG_THERMAL_GOV_STEP_WISE=y` | Step-wise governor (required for `type = "passive"` trip points) |
| `CONFIG_THERMAL_HWMON=y` | Exposes temperature readings via `/sys/class/hwmon` for monitoring tools |
| `CONFIG_SUN8I_THERMAL=y` | Allwinner sun8i/sun20i THS driver |

These are explicitly set even though `multi_v7_defconfig` enables them by default, to document the dependency and protect against future size-reduction changes inadvertently disabling them.

## Why No CPU Throttling

`CONFIG_CPU_THERMAL` (CPU frequency cooling) is **intentionally not enabled**. The DTS thermal zone has no `cooling-maps`, so no cooling device is bound to the 85°C passive trip — it fires as a monitoring event only. Enabling CPU frequency throttling would risk breaking Klipper's real-time stepper motor timing under sustained thermal load. Critical shutdown at 100°C is the safety backstop.

## Testing

After this PR, CPU temperature is readable via:
```bash
cat /sys/class/thermal/thermal_zone0/temp
# e.g. 42000 = 42.0°C

cat /sys/class/hwmon/hwmon0/temp1_input
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,4 @@
 /.vscode
+
+# Yocto build artifacts (generated locally)
+/build
--- a/meta-opencentauri/recipes-apps/klipper/files/mainboard.cfg
+++ b/meta-opencentauri/recipes-apps/klipper/files/mainboard.cfg
@@ -114,7 +114,7 @@ pin: PG17
 [fan_generic box_fan]
 pin: PG18
 
-[temperature_fan Host]
+[temperature_fan mainboard]
 pin: PG16
 tachometer_pin: PG6
 min_temp: 5
--- a/meta-opencentauri/recipes-apps/klipper/files/mainboard.cfg
+++ b/meta-opencentauri/recipes-apps/klipper/files/mainboard.cfg
@@ -114,7 +114,15 @@ pin: PG17
 [fan_generic box_fan]
 pin: PG18
 
-[controller_fan mainboard]
+[temperature_fan Host]
 pin: PG16
 tachometer_pin: PG6
-idle_speed: 0
+min_temp: 5
+max_temp: 85
+target_temp: 65
+min_speed: 0
+sensor_type: temperature_host
+control: pid
+pid_Kp: 2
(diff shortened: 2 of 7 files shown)


jamesturton (approved):
LGTM!

PR #99, opened 2026-04-06
Replace nginx with httpd

jamesturton:
Build busybox with httpd enabled. httpd is much lighter than nginx and saves roughly 1 Mb ram in our use case. We are not building httpd with reverse proxy support so moonraker and mainsail configs must be changed to use the correct port numbers.

Tested working on my machine


meta-opencentauri/recipes-apps/ustreamer/files/ustreamer-init-d
```diff
@@ -9,7 +9,7 @@
 ### END INIT INFO
 
 USTREAMER_EXEC="/usr/bin/ustreamer-wrapper"
-USTREAMER_ARGS="--device=/dev/video0 -r 864x480 -c HW -m MJPEG -b 1 -w 1 -f 10"
+USTREAMER_ARGS="-r 864x480 -c HW -m MJPEG -b 1 -w 1 -f 10 -l --host=0.0.0.0 --allow-origin=*"
```
jamesturton, line 12: From ustreamer docs:
```
    -l|--slowdown  ─────────────────────── Slowdown capturing to 1 FPS or less when no stream or sink clients
                                           are connected. Useful to reduce CPU consumption. Default: disabled.
```
Minimizing the camera feed in mainsail disconnects the feed, so this is a cheap win.
Also `/dev/video0` is default device so this is not needed.

jamesturton (commented):
Should also work if we switch back to mjpg-streamer although not tested.
Probably need to update moonraker port in fluidd too.

meta-opencentauri/recipes-apps/ustreamer/files/ustreamer-init-d
suchmememanyskill: @jamesturton Although, one thing i'm unsure about. I did test this, but saw no noticable difference in resource usage between when i was looking at a camera stream and when i was not. 

What is the default behaviour? Push no frames? Push max frames?

jamesturton: You are right. This only effects the capture rate, but as we aren't doing any re-coding it doesn't have any effect on cpu. And we still need the buffer in ram so it doesn't change memory usage.

meta-opencentauri/recipes-apps/fluidd/fluidd_1.36.2.bb
suchmememanyskill, line 34: Does fluidd not need a smilar config as mainsail to let it know where moonraker is located?

meta-opencentauri/recipes-httpd/nginx/nginx_%.bbappend
```diff
@@ -1,6 +0,0 @@
-FILESEXTRAPATHS:prepend := "${THISDIR}/${PN}:"
```
suchmememanyskill, line 1: Is it worth deleting this if we ever want to switch back?

meta-opencentauri/recipes-data/config-switcher/files/webui-nginx
```diff
@@ -1,56 +0,0 @@
-map $http_upgrade $connection_upgrade {
```
suchmememanyskill, line 1: Same for this, maybe worth keeping around?

meta-opencentauri/recipes-apps/fluidd/fluidd_1.36.2.bb
suchmememanyskill: It does not

meta-opencentauri/recipes-httpd/nginx/nginx_%.bbappend
jamesturton: I've kept this file just in case

meta-opencentauri/recipes-data/config-switcher/files/webui-nginx
jamesturton: Same here

PR #121, opened 2026-04-09
Remove httpd, serve mainsail direct from moonraker

jamesturton:
Saves an additional 1mb ram
Users don't have to use port 7125 for moonraker anymore


Atomique13 (approved):
Looks extremely good to me! MERGING!

PR #127, opened 2026-04-11
Correct second wipe location in WIPE_NOZZLE macro

BertrandTheHealer:
The M729 macro correctly wipes the nozzle between X173 and X187 where the brush/scrubber is. However in the WIPE_NOZZLE macro the second, slower wipe after touching off on the bed wipes between X110 and X140, where there is nothing to wipe on.

This corrects the second wipe to be in the same location as the first.

I have tested this config on my CC and verified it works as expected.


suchmememanyskill:
No longer relevant due to macros PR

PR #130, opened 2026-04-11
Add M8213 LED On and M8212 LED Off macros for compatibility with the patched firmware gcode

BertrandTheHealer:
The modified start and end gcode at https://docs.opencentauri.cc/patched-firmware/modified_start_end_machine_gcode/ use M8213 and M8212 gcode for turning on and off the chamber light. Those are not currently implemented in COSMOS so users will see an error if they keep the same gcode they were using for the patched firmware.

This adds the commands back to macros.cfg as aliases to the SET_LED function for compatibility.

### DIFF
--- a/meta-opencentauri/recipes-apps/klipper/files/macros.cfg
+++ b/meta-opencentauri/recipes-apps/klipper/files/macros.cfg
@@ -5,11 +5,13 @@
 
 # Turn on chamber LED
 [gcode_macro M8213]
-SET_LED LED=case WHITE=1.00 SYNC=0 TRANSMIT=1
+gcode:
+    SET_LED LED=case WHITE=1.00 SYNC=0 TRANSMIT=1
 
 # Turn off chamber LED
 [gcode_macro M8212]
-SET_LED LED=case WHITE=0.00 SYNC=0 TRANSMIT=1
+gcode:
+    SET_LED LED=case WHITE=0.00 SYNC=0 TRANSMIT=1
 
 [gcode_macro _COSMOS_SETTINGS]
 variable_sync_camera_led_to_chamber_led: False
--- a/meta-opencentauri/recipes-apps/klipper/files/macros.cfg
+++ b/meta-opencentauri/recipes-apps/klipper/files/macros.cfg
@@ -7,11 +7,13 @@
 [gcode_macro M8213]
 gcode:
     SET_LED LED=case WHITE=1.00 SYNC=0 TRANSMIT=1
+    SET_LED LED=hotend WHITE=1.00 SYNC=0 TRANSMIT=1
 
 # Turn off chamber LED
 [gcode_macro M8212]
 gcode:
     SET_LED LED=case WHITE=0.00 SYNC=0 TRANSMIT=1
+    SET_LED LED=hotend WHITE=1.00 SYNC=0 TRANSMIT=1
 
 [gcode_macro _COSMOS_SETTINGS]
 variable_sync_camera_led_to_chamber_led: False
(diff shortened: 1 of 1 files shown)


jamesturton:
The docs you linked specifically refer to the patched firmware, not cosmos. Reusing the same profile in orca slicer is not advised as you should be using the octo/klipper machine type and not elegoolink.

BertrandTheHealer:
> The docs you linked specifically refer to the patched firmware, not cosmos. Reusing the same profile in orca slicer is not advised as you should be using the octo/klipper machine type and not elegoolink.

I am using OrcaSlicer. I was previously using the patched firmware and when switching to COSMOS I did not change the printer profile, which contains the start gcode and machine parameters, only the "physical printer" profile which contains the connection settings and is located in a completely separate dialog. This seemed like the natural thing to do, and I would have lost or had to manually migrate other custom settings if I had switched to a completely different printer profile. Multiple people in the discord reported seeing errors in the print console that M8213 and M8212 are unknown commands after switching to COSMOS so I assume they did the same thing.

In my opinion this eliminates a potential pain point for users switching from patched firmware to COSMOS, and worst case doesn't hurt anything.

RowanSkie:
I seem to remember some offhand comment that these two gcodes will be re-added in Cosmos.

BertrandTheHealer:
> I seem to remember some offhand comment that these two gcodes will be re-added in Cosmos.

Ya that's what this PR does, if it's been done elsewhere let me know and I'll close it.

meta-opencentauri/recipes-apps/klipper/files/macros.cfg
```diff
@@ -3,6 +3,18 @@
 # EDITING THIS FILE WILL EXCLUDE YOUR CHANGES FROM FUTURE UPDATES.
 # -----
 
+# Turn on chamber LED
+[gcode_macro M8213]
+gcode:
+    SET_LED LED=case WHITE=1.00 SYNC=0 TRANSMIT=1
+    SET_LED LED=hotend WHITE=1.00 SYNC=0 TRANSMIT=1
```
suchmememanyskill, line 10: Instead of the hotend command: RUN_SHELL_COMMAND CMD=CAMERA_LIGHT_ON

meta-opencentauri/recipes-apps/klipper/files/macros.cfg
```diff
@@ -3,6 +3,18 @@
...
+
+# Turn off chamber LED
+[gcode_macro M8212]
+gcode:
+    SET_LED LED=case WHITE=0.00 SYNC=0 TRANSMIT=1
+    SET_LED LED=hotend WHITE=1.00 SYNC=0 TRANSMIT=1
```
suchmememanyskill, line 16: Instead of the hotend command: RUN_SHELL_COMMAND CMD=CAMERA_LIGHT_OFF

suchmememanyskill (changes requested):
M8212/M8213 never set the hotend LED, only the case LED. Probably worth keeping that behaviour.

suchmememanyskill:
One thing i'm unsure of, is it worth keeping the original gcode start/end script working?

Atom is working on a complete replacement for the start/end gcode, but this is not done get.

I guess it's not too much effort and works fine

Atomique13:
> One thing i'm unsure of, is it worth keeping the original gcode start/end script working?
> 
> Atom is working on a complete replacement for the start/end gcode, but this is not done get.
> 
> I guess it's not too much effort and works fine

My opinion is that users should have to choose between 2 sets of macros, one for the old patched setup and one for the new Cosmos. Maintaining both would quickly become a headache. My two cents are that it is better to adapt to the new Cosmos, since it aligns more closely with Klipper.

suchmememanyskill:
This is no longer relevant as patched firmware gcode isn't compatible anymore with COSMOS

PR #146, opened 2026-04-15
Add admonition about plugins

robertsamples:
warn users not to add plugins due to resource limitaiton


jamesturton:
Thank you for this!

PR #152, opened 2026-04-18
Add filament sensor configuration

boetti-dev:
- Add [filament_switch_sensor] section to machine.cfg
- Uses PC0 pin on main MCU (R528/Allwinner)
- Trigger level: LOW (active when filament is missing)
- Pauses print on filament runout using PAUSE macro

PAUSE macro has to be done properly


suchmememanyskill:
Fixed via macros PR

PR #155, opened 2026-04-23
Update machine configuration and macros and other stuff

Atomique13:
Added new configuration options for printer settings, including kinematics and acceleration parameters. Updated extruder and stepper motor settings for improved performance.


meta-opencentauri/recipes-apps/klipper/files/saved_variables.cfg
```diff
@@ -0,0 +1,12 @@
+[variables]
```
suchmememanyskill: Before it is forgotten:

This will not work, we have no way to update this file in the future if we allow users to edit it.

I'm also unsure yet if yet another configuration is the way to go

meta-opencentauri/recipes-apps/klipper/files/macros.cfg
```diff
@@ -3,19 +3,262 @@
 # EDITING THIS FILE WILL EXCLUDE YOUR CHANGES FROM FUTURE UPDATES.
 # -----
 
+
+
 [gcode_macro _COSMOS_SETTINGS]
```
suchmememanyskill: Should we maybe move these as well into the new variables file, however we do it?

Atomique13:
machine.cfg should be ready to use as it is now, macros....not so much, a lot of broken things there.

meta-opencentauri/recipes-apps/klipper/files/klipper-init-d
suchmememanyskill: Needs to be bumped to 0.0.7

meta-opencentauri/recipes-apps/klipper/files/machine.cfg
```diff
@@ -228,77 +294,40 @@ sclk_pin: bed:PC9
...
+[filament_switch_sensor filament_sensor]
+switch_pin: PC0
+pause_on_runout: false
+debounce_delay: 1
+runout_distance: 800
+runout_gcode:
+  RESPOND TYPE=echo MSG="Refill filament"
+  PAUSE STATE=runout
+immediate_runout_gcode:
+  RESPOND TYPE=echo MSG="Ran out of filament"
```
suchmememanyskill: Might be worth specifying that it'll use the last bit of filament

Something like "Ran out of filament, using remaining filament in PTFE"

meta-opencentauri/recipes-apps/klipper/files/machine.cfg
```diff
@@ -228,77 +294,40 @@ sclk_pin: bed:PC9
...
-screw2_name: front right screw
-screw3: 228.5, 228.5
-screw3_name: rear right screw
-screw4: 28.5, 228.5
-screw4_name: rear left screw
-horizontal_move_z: 5
-speed: 350
-screw_thread: CW-M3
+[led hotend]
+white_pin: hotend:PC9
```
suchmememanyskill, line 287: I thought you wanted to move this to an extras file?

meta-opencentauri/recipes-apps/klipper/files/machine.cfg
```diff
@@ -73,132 +140,131 @@ sensor_pin: hotend:PA3
...
+resistance1: 9206.431
+temperature2: 95
+resistance2: 4186.247
+temperature3: 293
+resistance3: 47.387
+
+[heater_fan extruder]
+pin: hotend:PC8
+tachometer_pin: hotend:PA1
+tachometer_poll_interval: 0.001
```
suchmememanyskill, line 174: What does this fix?

meta-opencentauri/recipes-apps/klipper/files/machine.cfg
```diff
@@ -73,132 +140,131 @@ sensor_pin: hotend:PA3
...
+sensors: hx71x sg0, hx71x sg1, hx71x sg2, hx71x sg3
+samples: 2
+samples_tolerance_retries: 3
+speed: 2
+sample_retract_dist: 1
+lift_speed: 5
+counts_per_gram: 29.11
+reference_tare_counts: 71000
+z_offset: -0.15
+samples_result: median
```
suchmememanyskill, line 235: is median a good idea on a probe setup as inaccurate as the cc?

meta-opencentauri/recipes-apps/klipper/files/machine.cfg
```diff
@@ -73,132 +140,131 @@ sensor_pin: hotend:PA3
...
+trigger_force: 50
+drop_first_result: true
```
suchmememanyskill, line 237: What does this fix?

meta-opencentauri/recipes-apps/klipper/files/machine.cfg
```diff
@@ -228,77 +294,40 @@ sclk_pin: bed:PC9
...
-z_offset: -0.15
-trigger_force: 50
-
-[printer]
-kinematics: corexy
-max_velocity: 500.000000
-max_accel: 20000.000000
-max_z_velocity: 20.000000
-max_z_accel: 500.000000
-square_corner_velocity: 9.000000
```
suchmememanyskill, line 250: square_corner_velocity now defaults to 5, did you see bad effects at 9mm/s corner speeds?

meta-opencentauri/recipes-apps/klipper/files/machine.cfg
```diff
@@ -73,132 +140,131 @@ sensor_pin: hotend:PA3
...
+damping_ratio_x: 0.06
+damping_ratio_y: 0.082
+
+[resonance_tester]
+probe_points: 128,128,20
+accel_chip: lis2dw
+min_freq: 10
+max_freq: 100
+accel_per_hz: 100
+hz_per_sec: 2
```
suchmememanyskill, line 197: Just to make sure, does this cut down on the number of data points? from 70 to 45?

meta-opencentauri/recipes-apps/klipper/files/machine.cfg
```diff
@@ -27,42 +58,78 @@ enable_pin: !PE10
...
 dir_pin: !PE13
 enable_pin: !PE15
 endstop_pin: tmc2209_stepper_y:virtual_endstop
 microsteps: 16
 full_steps_per_rotation: 200
-position_endstop: -2.5
-position_max: 270.000000
-position_min: -2.5
+position_endstop: -2
+position_max: 265
```
suchmememanyskill, line 85: How was elegoo off by 5mm? 😭

meta-opencentauri/recipes-apps/klipper/files/machine.cfg
```diff
@@ -27,42 +58,78 @@ enable_pin: !PE10
...
+rotation_distance: 8
 microsteps: 16
 full_steps_per_rotation: 200
-position_endstop: 4.5
-position_max: 258.000000
-position_min: -5
-homing_speed: 8.000000
+position_endstop: 4
+position_max: 258
+position_min: -4
```
suchmememanyskill: Wouldn't this be a goodn't idea, like wouldn't this scratch some people's plates where it didn't before?

meta-opencentauri/recipes-apps/klipper/files/machine.cfg
```diff
@@ -307,3 +336,30 @@ sensor_mcu: hotend
...
+tachometer_pin: PG6
+min_temp: 5
+max_temp: 85
+target_temp: 60
+min_speed: 0
+sensor_type: temperature_host
+control: pid
+pid_Kp: 2
+pid_Ki: 5
+pid_Kd: 0.5
```
suchmememanyskill, line 322: Didn't you want to remove this and just turn it on when the steppers are on?

meta-opencentauri/recipes-apps/klipper/files/machine.cfg
```diff
@@ -73,132 +140,131 @@ sensor_pin: hotend:PA3
...
+# CC1 CANVAS hotend config
+# Fixed beta value based on Atom's measurements
+# beta: 4800
+[thermistor elegoo_canvas_thermistor]
+temperature1: 72
+resistance1: 9206.431
+temperature2: 95
+resistance2: 4186.247
+temperature3: 293
+resistance3: 47.387
```
suchmememanyskill: Why aren't we using these over the beta values?
I guess it isn't a super big issue if the beta values aren't that far off

meta-opencentauri/recipes-apps/klipper/files/macros.cfg
```diff
@@ -3,252 +3,373 @@
...
+[gcode_macro _global_var]
+variable_target_extruder: 0
+variable_pause_park:{'x': 10, 'y': 240, 'z': 10, 'e': 2}
+variable_cancel_park:{'x': 240, 'y': 240, 'z': 10, 'e': 5}
+variable_extruder_temp:{'probing': 140, 'cleaning': 180, 'loading': 230, 'purging': 250}
+variable_z_maximum_lifting_distance: 250
+variable_pause_resume_travel_speed: 150
+variable_heatsoak: 0
+variable_adaptive_mesh: False
+variable_adaptive_purge: False
```
suchmememanyskill: Why is this false by default? What does it do when it's false?

meta-opencentauri/recipes-apps/klipper/files/macros.cfg
```diff
@@ -3,252 +3,373 @@
...
-variable_extruder_target: 230 
+[gcode_macro _global_var]
+variable_target_extruder: 0
+variable_pause_park:{'x': 10, 'y': 240, 'z': 10, 'e': 2}
+variable_cancel_park:{'x': 240, 'y': 240, 'z': 10, 'e': 5}
+variable_extruder_temp:{'probing': 140, 'cleaning': 180, 'loading': 230, 'purging': 250}
+variable_z_maximum_lifting_distance: 250
+variable_pause_resume_travel_speed: 150
+variable_heatsoak: 0
+variable_adaptive_mesh: False
```
suchmememanyskill: variable_adaptive_mesh and variable_adaptive_purge might be worth bringing over to cosmos.conf

meta-opencentauri/recipes-apps/klipper/files/macros.cfg
```diff
@@ -3,252 +3,373 @@
...
+variable_adaptive_purge: False
 gcode:
 
-[gcode_macro PRINT_START]
-gcode:
-    SET_GCODE_VARIABLE MACRO=_GLOBAL_VARS VARIABLE=extruder_target VALUE={params.EXTRUDER_TEMP}
 
-# Home Z with Strain Gauge, ends up with the nozzle 2mm above bed
-[gcode_macro STRAINGAUGE_Z_HOME]
+[gcode_macro PRINT_START]
```
suchmememanyskill, line 16: Could you add a check on
```
{% set all_calibrated = printer["gcode_macro _CHECK_CALIBRATION_VARS"].all_calibrated %}

{% if not all_calibrated %}
    SET_DISPLAY_TEXT MSG="Calibration process not completed. Please complete initial calibration first"
    CANCEL_PRINT
    RETURN
{% endif %}
```

meta-opencentauri/recipes-apps/klipper/files/macros.cfg
```diff
@@ -3,252 +3,373 @@
...
+    CLEAR_PAUSE
     G90
+    SET_DISPLAY_TEXT MSG="Bed: {target_bed}c"
+    M140 S{target_bed}
+    G28
+    {% if sensor.enabled and not sensor.filament_detected %}
+        SET_DISPLAY_TEXT MSG="NO FILAMENT LOADED!"
+        CANCEL_PRINT
+        RETURN
+    {% endif %}
```
suchmememanyskill, line 42: This check should probably be above line 29

meta-opencentauri/recipes-apps/klipper/files/macros.cfg
```diff
@@ -3,252 +3,373 @@
...
-
-    M140 S{bed_temp}                ; Set bed temp
-    M104 S140                       ; Set nozzle temp to 140°C
-
-    SET_GCODE_OFFSET Z=0 MOVE=0     ; Reset Z offset
-
-    G28 Z0                          ; Home Z only using photoelectric sensor
+    {% set svar = printer['gcode_macro _global_var'] %}
+    BED_MESH_CLEAR
+    M104 S{svar.extruder_temp.cleaning}
```
suchmememanyskill: Why are you cooling down when cleaning the nozzle?
For AFC this might be problematic

suchmememanyskill: Oh wait for AFC i'll just override this so won't be an issue there, but i don't understand the temperature swapping still

meta-opencentauri/recipes-apps/klipper/files/macros.cfg
```diff
@@ -3,252 +3,373 @@
...
 gcode:
 
-[gcode_macro PRINT_START]
-gcode:
-    SET_GCODE_VARIABLE MACRO=_GLOBAL_VARS VARIABLE=extruder_target VALUE={params.EXTRUDER_TEMP}
 
-# Home Z with Strain Gauge, ends up with the nozzle 2mm above bed
-[gcode_macro STRAINGAUGE_Z_HOME]
+[gcode_macro PRINT_START]
 gcode:
```
suchmememanyskill, line 17: How should AFC tool switching work? Orca probably should pass the tool index to PRINT_START

meta-opencentauri/recipes-apps/klipper/files/macros.cfg
```diff
@@ -3,252 +3,373 @@
...
+  # rehome after thermal expansion
+    SET_DISPLAY_TEXT MSG="Heatsoak: {svar.heatsoak}m"
+    G4 P{svar.heatsoak*1000*60}
+    LOADCELL_Z_HOME
+    {% if svar.adaptive_mesh %}
+        M109 S{svar.extruder_temp.probing}
+        SET_DISPLAY_TEXT MSG="Bed mesh adaptive"
+        BED_MESH_CALIBRATE ADAPTIVE=1
+    {% else %}
+        {% if "default" in printer.bed_mesh.profiles %}
```
suchmememanyskill, line 85: This implicitly removes a feature before where you could change the bed mesh name. Can you maybe make this a (global) var?

meta-opencentauri/recipes-apps/klipper/files/macros.cfg
suchmememanyskill: It seems like there's no purge line. Would it be possible to restore default behaviour on false? (line at the front of the plate?)

meta-opencentauri/recipes-apps/klipper/files/macros.cfg
```diff
@@ -3,252 +3,373 @@
...
+        SMART_PARK
+    {% endif %}
 
-    M400                            ; Wait for moves to finish
+    SET_DISPLAY_TEXT MSG="Hotend: {target_extruder}c"
+    M109 S{target_extruder}
+    {% if svar.adaptive_purge %}
+        LINE_PURGE
+    {% endif %}
+    SET_DISPLAY_TEXT MSG="Printer goes brr"
```
suchmememanyskill: Would prefer to not set such messages.
Could the message instead be cleared here (so there isn't a persistent banner text)

meta-opencentauri/recipes-apps/klipper/files/macros.cfg
```diff
@@ -3,252 +3,373 @@
...
 
-    # Move up 2mm from the probed position
-    G91                             ; Relative positioning
-    G1 Z2 F600                      ; Move up 2mm
-    M400                            ; Wait for moves to finish
-    G90                             ; Absolute positioning
+        {% if (printer.gcode_move.position.z + 10) < z_max %}
+            G1 Z+10 F3000
+        {% else %}
+            G1 Z+{(z_max - printer.gcode_move.position.z)} F3000
```
suchmememanyskill, line 135: nit: Can you not do 

```
G90
G1 Z{z_max} F3000
```

Also can't wait for this to fail as there's trash at the bottom of the printer

meta-opencentauri/recipes-apps/klipper/files/macros.cfg
```diff
@@ -3,252 +3,373 @@
...
+        G90
+        G1 X240 Y240 F9000
+    {% endif %}
+    ALL_FANS_OFF
+    TURN_OFF_HEATERS 
+    M220 S100
+    M221 S100
+    CLEAR_PAUSE
+    M84
+    SET_DISPLAY_TEXT MSG="Finish Print!"
```
suchmememanyskill: I don't see this as a very useful message, as mainsail says the same thing already

meta-opencentauri/recipes-apps/klipper/files/macros.cfg
```diff
@@ -3,252 +3,373 @@
...
+        {% if (printer.gcode_move.position.z + 10) < z_max %}
+            G1 Z+10 F3000
+        {% else %}
+            G1 Z+{(z_max - printer.gcode_move.position.z)} F3000
+        {% endif %}
+        G90
+        G1 X240 Y240 F9000
+    {% endif %}
+    ALL_FANS_OFF
+    TURN_OFF_HEATERS
```
suchmememanyskill: Might be worth keeping naming consistent between internal klipper macros and ours

like not ALL_FANS_OFF, but TURN_OFF_FANS

meta-opencentauri/recipes-apps/klipper/files/macros.cfg
```diff
@@ -3,252 +3,373 @@
...
+[gcode_macro CANCEL_PRINT]
+description: 
+rename_existing: CANCEL_PRINT_BASE
+gcode:
+    {% set svar = printer['gcode_macro _global_var'] %}
+    {% set x_park = svar.cancel_park.x %}
+    {% set y_park = svar.cancel_park.y %}
+    {% set z_park = svar.cancel_park.z %}
+    {% set z_lift_max = svar.z_maximum_lifting_distance %}
+    {% set e_restract = svar.cancel_park.e %}
```
suchmememanyskill: e_retract?

meta-opencentauri/recipes-apps/klipper/files/macros.cfg
```diff
@@ -3,252 +3,373 @@
...
+        {% if sensor.enabled and sensor.filament_detected %}
+            {% if (printer.extruder.target != 0 and printer.extruder.temperature >= printer.extruder.target and printer.extruder.temperature >= e_mintemp) %}
+                G1 E-{e_restract} F2700
+            {% else %}
+                RESPOND TYPE=command MSG="Nozzle not hot enough"
+            {% endif %}
+        {% endif %}
 
-    # TODO: Restore bed mesh
+        {%if (printer.gcode_move.position.z + 10) < z_lift_max %}
```
suchmememanyskill: missing space before if

meta-opencentauri/recipes-apps/klipper/files/macros.cfg
```diff
@@ -3,252 +3,373 @@
...
+            G1 Z+10 F3000
+        {% else %}
+            G1 Z+{(z_lift_max - printer.gcode_move.position.z)} F3000
+        {% endif %}
+            G90
+            G1 X{x_park} Y{y_park} F9000
```
suchmememanyskill: bad indentation

meta-opencentauri/recipes-apps/klipper/files/macros.cfg
```diff
@@ -3,252 +3,373 @@
...
+    {% set sensor = printer['filament_switch_sensor filament_sensor'] %}
+
+    CANCEL_PRINT_BASE
+    {% if printer.toolhead.homed_axes|lower == "xyz" %}
+        G91
+        {% if sensor.enabled and sensor.filament_detected %}
+            {% if (printer.extruder.target != 0 and printer.extruder.temperature >= printer.extruder.target and printer.extruder.temperature >= e_mintemp) %}
+                G1 E-{e_restract} F2700
+            {% else %}
+                RESPOND TYPE=command MSG="Nozzle not hot enough"
```
suchmememanyskill: type command?

meta-opencentauri/recipes-apps/klipper/files/macros.cfg
```diff
@@ -3,252 +3,373 @@
...
+            G1 X{x_park} Y{y_park} F9000
+    {% endif %}
+    ALL_FANS_OFF
+    TURN_OFF_HEATERS 
+    M220 S100
+    M221 S100
+    CLEAR_PAUSE
+    M84
+    SET_GCODE_VARIABLE MACRO=_global_var VARIABLE=target_extruder VALUE=0
+    SET_DISPLAY_TEXT MSG="Cancel Print Success!"
```
suchmememanyskill: This to me sounds more like a task failed successfully kinda deal.
I don't think we need a display text message here. Mainsail will inform you when execution halts

meta-opencentauri/recipes-apps/klipper/files/macros.cfg
```diff
@@ -3,252 +3,373 @@
 # EDITING THIS FILE WILL EXCLUDE YOUR CHANGES FROM FUTURE UPDATES.
 # -----
 
-[gcode_macro _GLOBAL_VARS]
-variable_extruder_target: 230 
+[gcode_macro _global_var]
+variable_target_extruder: 0
+variable_pause_park:{'x': 10, 'y': 240, 'z': 10, 'e': 2}
```
suchmememanyskill: This doesn't sound like a good idea, as this makes for an unhappy cable me thinks.
Why is this different than the cancel position?
Why can't this just be the chute park?

meta-opencentauri/recipes-apps/klipper/files/macros.cfg
```diff
@@ -3,252 +3,373 @@
...
+    {% set svar = printer['gcode_macro _global_var'] %}
+    {% set state = params.STATE|default('normal') %}
+    {% set x_park = svar.pause_park.x|float %}
+    {% set y_park = svar.pause_park.y|float %}
+    {% set e_retract = svar.pause_park.e|float %}
+    {% set z_lift_max = svar.z_maximum_lifting_distance|int %}
+    {% set sensor = printer['filament_switch_sensor filament_sensor'] %}
+    {% set pos = printer.gcode_move.position %}
+
+    {% if printer.pause_resume.is_paused == False %}
```
suchmememanyskill: Can you invert this check and RETURN instead? That way you don't indent as much

meta-opencentauri/recipes-apps/klipper/files/macros.cfg
```diff
@@ -3,252 +3,373 @@
...
+        SET_DISPLAY_TEXT MSG="Pausing"
+
+        PAUSE_BASE
+
+
+        {% if (printer.extruder.temperature + 5 >= printer.extruder.target) and printer.extruder.can_extrude  %}
+            {% if sensor.enabled and sensor.filament_detected %}
```
suchmememanyskill: nit: nested if

meta-opencentauri/recipes-apps/klipper/files/macros.cfg
```diff
@@ -3,252 +3,373 @@
...
+                G91
```
suchmememanyskill: Prefer M83

meta-opencentauri/recipes-apps/klipper/files/macros.cfg
```diff
@@ -3,252 +3,373 @@
...
+        {% endif %}
 
-    RESTORE_GCODE_STATE NAME=straingauge_z_home_state
+        {% if printer.toolhead.homed_axes|lower == "xyz" %}
+            {% set zpos = printer.gcode_move.position.z %}
+            G91
+            {% if (zpos + 5) < z_lift_max %}
+                G1 Z+5 F3000
+            {% else %}
+                G1 Z+{(z_lift_max - zpos)} F3000
```
suchmememanyskill: same comment as earlier

meta-opencentauri/recipes-apps/klipper/files/macros.cfg
```diff
@@ -3,252 +3,373 @@
...
+            {% endif %}
+            G90
+            {% if pos.x != x_park or pos.y != y_park %}
```
suchmememanyskill: This check is redundant, klipper immediately finishes a move if it's already at the correct position

meta-opencentauri/recipes-apps/klipper/files/macros.cfg
```diff
@@ -3,252 +3,373 @@
...
+    CLEAR_PAUSE
+    M84
+    SET_GCODE_VARIABLE MACRO=_global_var VARIABLE=target_extruder VALUE=0
+    SET_DISPLAY_TEXT MSG="Cancel Print Success!"
+
+[gcode_macro PAUSE]
+rename_existing: PAUSE_BASE
+variable_state: 'normal'
+gcode:
+    {% set svar = printer['gcode_macro _global_var'] %}
```
suchmememanyskill: nit: i think i'd prefer the var name global

meta-opencentauri/recipes-apps/klipper/files/macros.cfg
```diff
@@ -3,252 +3,373 @@
...
-# Wipe Nozzle, see auto_leveling.cpp:443
-[gcode_macro M729]
-gcode:
-    SAVE_GCODE_STATE NAME=m729_saved_state
+        # Keep heater on so resume can prime quickly
+        M104 S{svar.extruder_temp.probing}
 
-    G90
+        {% if state == 'normal' %}
+            SET_DISPLAY_TEXT MSG="Paused normal"
```
suchmememanyskill: I feel like a normal pause doesn't need a status message

meta-opencentauri/recipes-apps/klipper/files/macros.cfg
```diff
@@ -3,252 +3,373 @@
...
 
-    # Move to tray
-    G1 X202 F6000
-    G1 Y264.5 F6000
+        {% elif state == 'filament_change' %}
+            # Ensure nozzle is up to temp for filament change/unload
+            {% if printer.extruder.temperature + 5 < printer.extruder.target %}
+                M109 S{printer.extruder.target}
```
suchmememanyskill: I don't see why you wouldn't always run M109 Sxxx here

meta-opencentauri/recipes-apps/klipper/files/macros.cfg
```diff
@@ -3,252 +3,373 @@
...
 
-    # Wipe
-    G1 X187 F8000
-    G1 X173
-    G1 X187
-    G1 X173
-    G1 X187
-    G1 X173
+    {% else %}
+        SET_DISPLAY_TEXT MSG="Printer already paused"
```
suchmememanyskill: I don't think we need a permanent status message. Maybe a RESPOND would be better

meta-opencentauri/recipes-apps/klipper/files/macros.cfg
```diff
@@ -3,252 +3,373 @@
...
+    {% endif %}
 
-    # Add some clearance
-    G1 Y250
+[delayed_gcode _resume_wait]
+gcode:
+    {% if printer['gcode_macro RESUME'].execute|lower != 'false' %}
+        RESUME
+    {% endif %}
```
suchmememanyskill: What is this supposed to do?

meta-opencentauri/recipes-apps/klipper/files/macros.cfg
```diff
@@ -3,252 +3,373 @@
...
-    # Move back to tray
-    G1 X202
-    G1 Y264.5 F1200
+[gcode_macro RESUME]
+rename_existing: RESUME_BASE
+variable_state: 'normal'
+gcode:
+    {% set svar = printer['gcode_macro _global_var'] %}
+    {% set e_retract = svar.pause_park.e|float %}
+    #{% set extruder_target_temp = printer.extruder.target|int %}
```
suchmememanyskill: We have version control, commenting out code isn't super nessesary. Just remove it

meta-opencentauri/recipes-apps/klipper/files/macros.cfg
```diff
@@ -3,252 +3,373 @@
...
-    M400
+    # Ensure nozzle heated to target before attempting to prime/extrude
+    # if printer.extruder.temperature + 5 < extruder_target_temp %}
+         #SET_DISPLAY_TEXT MSG="Heating nozzle to {extruder_target_temp}C"
+    M109 S{extruder_target_temp}
+    # {% endif %}
+    G91
+    G1 E{e_retract} F300
+    G90
+    SET_DISPLAY_TEXT MSG="Resuming"
```
suchmememanyskill: Imo unnesary

meta-opencentauri/recipes-apps/klipper/files/macros.cfg
```diff
@@ -3,252 +3,373 @@
...
 
-    # Clear command queue
-    M400
+    # Ensure nozzle heated to target before attempting to prime/extrude
+    # if printer.extruder.temperature + 5 < extruder_target_temp %}
+         #SET_DISPLAY_TEXT MSG="Heating nozzle to {extruder_target_temp}C"
+    M109 S{extruder_target_temp}
+    # {% endif %}
+    G91
+    G1 E{e_retract} F300
```
suchmememanyskill: Is it worth maybe priming the nozzle? (Purge & wipe)
Or does the client.cfg do this already?

meta-opencentauri/recipes-apps/klipper/files/macros.cfg
```diff
@@ -3,252 +3,373 @@
...
+    {% if printer['gcode_macro RESUME'].execute|lower != 'false' %}
+        RESUME
+    {% endif %}
 
-    # Move back to tray
-    G1 X202
-    G1 Y264.5 F1200
+[gcode_macro RESUME]
+rename_existing: RESUME_BASE
+variable_state: 'normal'
```
suchmememanyskill: this seems unused

meta-opencentauri/recipes-apps/klipper/files/macros.cfg
```diff
@@ -3,252 +3,373 @@
...
+    M220 S100
+    M221 S100
+    CLEAR_PAUSE
+    M84
+    SET_GCODE_VARIABLE MACRO=_global_var VARIABLE=target_extruder VALUE=0
+    SET_DISPLAY_TEXT MSG="Cancel Print Success!"
+
+[gcode_macro PAUSE]
+rename_existing: PAUSE_BASE
+variable_state: 'normal'
```
suchmememanyskill: this seems unused

meta-opencentauri/recipes-apps/klipper/files/macros.cfg
```diff
@@ -3,252 +3,373 @@
...
+    {% if "xyz" not in printer.toolhead.homed_axes %}
+        G28                         ; Home if necessary
+    {% endif %}
     # Move to purge position
-    G1 X202 Y250 F20000             ; Rapid move to purge area
-    G1 Y264.5 F1200                 ; Move to exact purge Y position
-    M109 S250                       ; Heat nozzle to 250°C and wait
+    MOVE_TO_TRAY
+    M104 S{svar.extruder_temp.purging}                       ; Heat nozzle to purging temp
+    TEMPERATURE_WAIT SENSOR=extruder MINIMUM={svar.extruder_temp.purging} MAXIMUM={svar.extruder_temp.purging+5}
```
suchmememanyskill: Why not combine this into an M109?

meta-opencentauri/recipes-apps/klipper/files/macros.cfg
```diff
@@ -3,252 +3,373 @@
...
+        G1 E10 F150                      ; Extrude 2mm slowly (ensure flow)
+        M400                            ; Wait for completion
+    {% endfor %}
+    M104 S{svar.extruder_temp.loading}
+    TEMPERATURE_WAIT SENSOR=extruder MAXIMUM={svar.extruder_temp.loading} MINIMUM={svar.extruder_temp.loading-5}
+    {% for purge in range(4) %}
+        G1 E25 F300                     ; Extrude 13mm at medium speed
+        G1 E10 F150                      ; Extrude 2mm slowly (ensure flow)
+        M400                            ; Wait for completion
+    {% endfor %}
```
suchmememanyskill: why do we have 2 purge cycles at 2 different temperatures?

meta-opencentauri/recipes-apps/klipper/files/macros.cfg
```diff
@@ -3,252 +3,373 @@
...
-    G1 E13 F523                     ; Repeat purge cycle 6
-    G1 E2 F150
-    M400
-    G1 E13 F523                     ; Repeat purge cycle 7
-    G1 E2 F150
-    M400
-    G1 E13 F523                     ; Repeat purge cycle 8
-    G1 E2 F150
-    M400
+    M106 S0                       ; Turn on part cooling fan (full speed)
```
suchmememanyskill: Either wrong value or wrong comment

meta-opencentauri/recipes-apps/klipper/files/macros.cfg
```diff
@@ -3,252 +3,373 @@
...
+    {% for purge in range(4) %}
+        G1 E25 F300                     ; Extrude 13mm at medium speed
+        G1 E10 F150                      ; Extrude 2mm slowly (ensure flow)
+        M400                            ; Wait for completion
+    {% endfor %}
+    M104 S{svar.extruder_temp.loading}
```
suchmememanyskill: It might be a good idea to wipe inbetween so the filament can fall throught the chute instead of staying there

meta-opencentauri/recipes-apps/klipper/files/macros.cfg
```diff
@@ -3,252 +3,373 @@
...
+    M104 S0
+    M106 S255
+    TEMPERATURE_WAIT SENSOR=extruder MAXIMUM={svar.extruder_temp.cleaning} MINIMUM={svar.extruder_temp.cleaning-5}
+    M107
+    G0 X170
+    G0 Y250
+    G4 P1000
+    MOVE_TO_TRAY
+    G0 X170
+    G0 Y250
```
suchmememanyskill: No move happened, why are we moving then moving to tray?
Might be worth invoking the wipe macro instead

meta-opencentauri/recipes-apps/klipper/files/macros.cfg
```diff
@@ -3,252 +3,373 @@
...
-    BED_MESH_CALIBRATE              ; Start bed mesh calibration
+    # Clear command queue
 
-    MOVE_TO_TRAY                    ; Move to tray after calibration
-    M106 S0                         ; Turn off part cooling fan
-    SET_FAN_SPEED FAN=model_helper_fan SPEED=0 ; Turn off auxiliary fan
-    M104 S0                         ; Turn off nozzle heater
-    M140 S0                         ; Turn off bed heater
+    M107 
+    M400
```
suchmememanyskill, line 305: Please clear the status message at the end of this macro: `SET_DISPLAY_TEXT` (no args)

meta-opencentauri/recipes-apps/klipper/files/macros.cfg
```diff
@@ -3,252 +3,373 @@
...
 gcode:
-    _CLOSE_PROMPT                   ; Close any existing prompt
+                       ; Close any existing prompt
     M83                            ; Relative extrusion mode
     G92 E0                         ; Reset extruder position
-    G1 E120 F240                   ; Move filament into the extruder (120mm at 4mm/s)
-    M729                           ; Wipe nozzle after loading
+    G1 E70 F420
+                       ; Move filament into the extruder (120mm at 4mm/s)
+    G4 P1000
```
suchmememanyskill: why we waiting

meta-opencentauri/recipes-apps/klipper/files/macros.cfg
```diff
@@ -3,252 +3,373 @@
...
+    _CLOSE_PROMPT
+    #CLEAN_NOZZLE                           ; Wipe nozzle after loading
```
suchmememanyskill: again, commented out code isn't really nessesary with version control

meta-opencentauri/recipes-apps/klipper/files/macros.cfg
```diff
@@ -3,252 +3,373 @@
...
-    _CLOSE_PROMPT                   ; Close any existing prompt
+                       ; Close any existing prompt
     M83                            ; Relative extrusion mode
     G92 E0                         ; Reset extruder position
-    G1 E120 F240                   ; Move filament into the extruder (120mm at 4mm/s)
-    M729                           ; Wipe nozzle after loading
+    G1 E70 F420
+                       ; Move filament into the extruder (120mm at 4mm/s)
+    G4 P1000
+    _CLOSE_PROMPT
```
suchmememanyskill: _CLOSE_PROMPT should be at the start of the macro

meta-opencentauri/recipes-apps/klipper/files/macros.cfg
```diff
@@ -258,152 +379,159 @@ gcode:
...
 [gcode_macro _LOAD_FILAMENT_STEP_DONE]
 gcode:
-    M729                            ; Wipe nozzle after loading
-    M104 S0                         ; Turn off nozzle heater
-    M106 S0                         ; Turn off part cooling fan
     _CLOSE_PROMPT                    ; Close the loading prompt
-
+    CLEAN_NOZZLE                            ; Wipe nozzle after loading
+    M104 S0                         ; Turn off nozzle heater
+    #M106 S0                         ; Turn off part cooling fan
```
suchmememanyskill: same as earlier

meta-opencentauri/recipes-apps/klipper/files/macros.cfg
```diff
@@ -258,152 +379,159 @@ gcode:
...
+    {% if "xyz" not in printer.toolhead.homed_axes %}
+        G28                         ; Home if necessary
+    {% endif %}
 
     MOVE_TO_TRAY                    ; Move to the tray
-    M109 S{extruder_temp}           ; Wait for nozzle to reach temp
-    M106 S255                       ; Turn on part cooling fan (full speed)
-    _LOAD_FILAMENT_STEP_PUSH        ; Push filament into extruder and wipe
+    M104 S{svar.extruder_temp.loading}
+    TEMPERATURE_WAIT SENSOR=extruder MINIMUM={svar.extruder_temp.loading} MAXIMUM={svar.extruder_temp.loading+5}
```
suchmememanyskill: Again, unsure why not M109

meta-opencentauri/recipes-apps/klipper/files/macros.cfg
```diff
@@ -258,152 +379,159 @@ gcode:
...
+    M107                       ; Turn on part cooling fan (full speed)
```
suchmememanyskill: Comment does not match gcode (turn fan off)

meta-opencentauri/recipes-apps/klipper/files/macros.cfg
```diff
@@ -3,252 +3,373 @@
...
+    G0 Y250 F12000             ; Move near to the try
+    G0 Y265 F1200                 ; Move slowly back to the tray
 
     M400                            ; Wait for movements to complete
 
+
 [gcode_macro _LOAD_FILAMENT_STEP_PUSH]
 gcode:
-    _CLOSE_PROMPT                   ; Close any existing prompt
+                       ; Close any existing prompt
```
suchmememanyskill: Can you check temp here? It's possible the idle timeout hit before this macro gets executed

meta-opencentauri/recipes-apps/klipper/files/macros.cfg
```diff
@@ -258,152 +379,159 @@ gcode:
...
     {% endif %}
-
-    G90                             ; Absolute positioning
-    G0 Z14.5 F600                   ; Ensure we have some Z clearance
-    G0 Y30 F15000                    ; Get some clearance for Y
-    G0 X255 F8000                  ; Move to cutting position
-    G0 Y3 F1200                     ; Cut
-    G0 Y30 F8000                   ; Retract from cutting
+                      ; Absolute positioning
+    CUT_FILAMENT                            ; Wait for movements to complete
```
suchmememanyskill, line 426: CUT_FILAMENT is undefined

meta-opencentauri/recipes-apps/klipper/files/macros.cfg
```diff
@@ -258,152 +379,159 @@ gcode:
...
-    G28 X 
-    M400
-
-    # Move away
-    G91
-    G1 X-20 F4800
+    BED_MESH_CLEAR
     G90
+    G0 Z5 F1200 ; 5 up zhop
+    {% if not rawparams %}
```
suchmememanyskill, line 455: Why is G28 different than G28 XYZ?

meta-opencentauri/recipes-apps/klipper/files/macros.cfg
```diff
@@ -258,152 +379,159 @@ gcode:
...
+        G28 X
+        G28 Y
+        G28 X
+        G0 X128 Y128 F4800
+        G28 Z
+    {% else %}
```
suchmememanyskill, line 461: G28 XYZ will fail

meta-opencentauri/recipes-apps/klipper/files/macros.cfg
```diff
@@ -258,152 +379,159 @@ gcode:
...
-    {% set HOME_CURRENT_X = printer.configfile.settings['tmc2209 stepper_x'].home_current|float %}
-    {% set HOME_CURRENT_Y = printer.configfile.settings['tmc2209 stepper_y'].home_current|float %}
+    {% set svar = printer['gcode_macro _global_var'] %}
+    {% set bed_temp = params.BED_TEMP|default(60)|float %}
+    M140 S{bed_temp}
+    G28
+    CLEAN_NOZZLE
+    M190 S{bed_temp}
+    M104 S{svar.extruder_temp.probing}
+    G4 S60000
```
suchmememanyskill: Did you mean `G4 P60000`

meta-opencentauri/recipes-apps/klipper/files/macros.cfg
```diff
@@ -258,152 +379,159 @@ gcode:
...
-      {% endif %}
-      # Wait for the Z clearance move to finish before starting XY homing moves.
-      M400
-    {% endif %}
 
-    # First X home. Note this may crash into the back steppers.
-    {% if home_all or 'X' in params %}
-      _HOME_X
+[gcode_macro _IDLE_TIMEOUT]
+gcode:
```
suchmememanyskill, line 603: Not turning off heaters during a print kinda sounds like a bad idea.

Ok, well, the bed probably should stay on. The nozzle can cool down though.

Also, unsure if it's worth a display message. RESPOND is probably better here. I think the message should be consistent across branches too, something like "Idle timeout reached, turning off heaters"

meta-opencentauri/recipes-apps/klipper/files/macros.cfg
```diff
@@ -258,152 +379,159 @@ gcode:
...
-    # Second X home (if homing both X and Y) to home x at a known good position for sensorless homing
-    {% if home_all or ('X' in params and 'Y' in params) %}
-      _HOME_X
-    {% endif %}
+[gcode_macro ALL_FANS_OFF]
+gcode:
+    M107
+    SET_FAN_SPEED FAN=aux_fan SPEED=0
+    SET_FAN_SPEED FAN=case_fan SPEED=0
+    SET_DISPLAY_TEXT MSG="ALL FANS OFF!"
```
suchmememanyskill: Don't see the value in this display message

meta-opencentauri/recipes-apps/klipper/files/macros.cfg
```diff
@@ -432,19 +560,13 @@ gcode:
...
     SET_DISPLAY_TEXT_BASE MSG="{message}"
 
-[gcode_macro _CLIENT_VARIABLE]
-variable_use_custom_pos: True
-variable_custom_park_x: 202
-variable_custom_park_y: 264.5
+
+[gcode_macro M729]
 gcode:
+    SET_DISPLAY_TEXT MSG="USE COSMOS GCODE!"
```
suchmememanyskill: "Use COSMOS start/end machine gcode"

meta-opencentauri/recipes-apps/klipper/files/kamp.cfg
```diff
@@ -0,0 +1,351 @@
...
+
+        {% endif %}
+
+        RESTORE_GCODE_STATE NAME=Prepurge_State                                                 # Restore gcode state
+    
+    {% endif %}
+
+
+
+[gcode_macro ELEGOO_PURGE]
```
suchmememanyskill, line 149: This currently is unused it seems?

Is it worth using this in the adaptive purge = false branch?

meta-opencentauri/recipes-apps/klipper/files/klipper-init-d
Atomique13: Gonna do

meta-opencentauri/recipes-apps/klipper/files/machine.cfg
Atomique13: sure

meta-opencentauri/recipes-apps/klipper/files/machine.cfg
Atomique13: as per my last email we leave it as it is now

meta-opencentauri/recipes-apps/klipper/files/machine.cfg
Atomique13: yes, cause you almost always get 1 outlier that has no use, if you get 2 counts similar enough that's good enough, average would get bamboozled by the outlier.

meta-opencentauri/recipes-apps/klipper/files/machine.cfg
Atomique13: If you want I can put it back to 9, 5 is the lowest for most printers, 10 is average, 9 will always looks worse than 5, but no one done the comparisons to see how slow printer gets.

meta-opencentauri/recipes-apps/klipper/files/machine.cfg
Atomique13: yes, we need the curve shape not the points, also faster

meta-opencentauri/recipes-apps/klipper/files/machine.cfg
Atomique13: have no idea but 270 was hitting frame with toolhead, I checked twice

meta-opencentauri/recipes-apps/klipper/files/machine.cfg
Atomique13: This saves people’s build plates because it reduces how far the nozzle is forced downward into the plate. Most users only need around -2 mm to reach nozzle-to-plate contact, so allowing -4 mm already adds an extra 2 mm of unnecessary travel that can go wrong.

The current default allows up to -5 mm downward movement, which is excessive and can easily damage hardware.

It also helps expose mechanical bed issues. For example, one user somehow required -16 mm offset. After redoing the bed screws and fixing the mechanical alignment, the required offset dropped back to around -2 mm, which is within normal range

meta-opencentauri/recipes-apps/klipper/files/machine.cfg
Atomique13: good catch

meta-opencentauri/recipes-apps/klipper/files/machine.cfg
Atomique13: The beta model follows a generic NTC curve, so in most cases it works well enough with nearly any thermistor. Steinhart-Hart coefficients, on the other hand, define a custom curve tailored to a specific thermistor. In practice, the two are very similar, but Steinhart-Hart tends to be more accurate below 50 °C.

Personally, I would still use the beta model instead of Steinhart-Hart for two reasons:

We can define a minimum hotend temperature to detect whether the hotend has been removed from the CANVAS board. With the beta model, the temperature difference used for detection is around 15 °C, which gives a good safety margin. Steinhart-Hart reduces that gap to around 5 °C because it corrects the curve more accurately, and that smaller margin can lead to false thermistor failure detections

meta-opencentauri/recipes-apps/klipper/files/macros.cfg
Atomique13: per my last email we gonna set to yes and see how it goes

meta-opencentauri/recipes-apps/klipper/files/kamp.cfg
Atomique13: well users can change to it if they want to, or smth it is an easter egg

meta-opencentauri/recipes-apps/klipper/files/macros.cfg
Atomique13: noted

meta-opencentauri/recipes-apps/klipper/files/macros.cfg
Atomique13: users are aware that fans are off

meta-opencentauri/recipes-apps/klipper/files/macros.cfg
Atomique13: yes

meta-opencentauri/recipes-apps/klipper/files/machine.cfg
Atomique13: it is better, cause first result could be wrong or dirty nozzle so we use that as sacrifical nozzle bump, bambu does the same

meta-opencentauri/recipes-apps/klipper/files/macros.cfg
Atomique13: could altho dunno

meta-opencentauri/recipes-apps/klipper/files/macros.cfg
Atomique13: yes we could

meta-opencentauri/recipes-apps/klipper/files/macros.cfg
Atomique13: it won't

meta-opencentauri/recipes-apps/klipper/files/macros.cfg
Atomique13: maybe I want my own klipper

meta-opencentauri/recipes-apps/klipper/files/machine.cfg
Atomique13: fan rpm reading

meta-opencentauri/recipes-apps/klipper/files/macros.cfg
Atomique13: done

meta-opencentauri/recipes-apps/klipper/files/macros.cfg
Atomique13: yes it is the retract amount for pause park, users can edit it in global vars

meta-opencentauri/recipes-apps/klipper/files/macros.cfg
Atomique13: ty sims lint

meta-opencentauri/recipes-apps/klipper/files/macros.cfg
Atomique13: fuck you honestly, for a coder you really hate debug comments

meta-opencentauri/recipes-apps/klipper/files/macros.cfg
Atomique13: M109 waits for temp to stabilise, temperature wait exits waiting once condition is reached.

meta-opencentauri/recipes-apps/klipper/files/macros.cfg
Atomique13: wrong comment

meta-opencentauri/recipes-apps/klipper/files/macros.cfg
Atomique13: Cause maybe users want to home just x or just y or just z not all always, also that's a normal klipper behaviour, homing x shouldnt home all lmao

meta-opencentauri/recipes-apps/klipper/files/macros.cfg
Atomique13: it will not 😝

meta-opencentauri/recipes-apps/klipper/files/macros.cfg
Atomique13: yeah sorry AI sims

meta-opencentauri/recipes-apps/klipper/files/macros.cfg
Atomique13: m109 waits for temp to stabilise we don't need that here, wait just check is if in range then continues

meta-opencentauri/recipes-apps/klipper/files/macros.cfg
Atomique13: Bambu does the same, I can remove it if u want

meta-opencentauri/recipes-apps/klipper/files/macros.cfg
Atomique13: we don't have a wipe macro, this is a wipe macro, that just wipes, I wasn't gonna do a wipe macro for 2 commands

meta-opencentauri/recipes-apps/klipper/files/macros.cfg
Atomique13: Sure

meta-opencentauri/recipes-apps/klipper/files/macros.cfg
Atomique13: oopsie, added now

meta-opencentauri/recipes-apps/klipper/files/macros.cfg
Atomique13: removed

meta-opencentauri/recipes-apps/klipper/files/macros.cfg
Atomique13: removed

meta-opencentauri/recipes-apps/klipper/files/macros.cfg
Atomique13: removed

meta-opencentauri/recipes-apps/klipper/files/macros.cfg
Atomique13: fixed

meta-opencentauri/recipes-apps/klipper/files/macros.cfg
Atomique13: should be easier for user to know what happened if toolhead is left or right, toolhead has to be at the back for pause to let users do things on prints if needed (add magnets dunno). If you want I can put both on the back right, I avoid using chute, chute can't home and or can hit things if movement is not accounting for stepper/lever which klipper isn't aware off.

meta-opencentauri/recipes-apps/klipper/files/macros.cfg
Atomique13: I hate this so much, default is default, not whatever they name it

meta-opencentauri/recipes-apps/klipper/files/machine.cfg
Atomique13: fixed

meta-opencentauri/recipes-apps/klipper/files/macros.cfg
Atomique13: fixed

meta-opencentauri/recipes-apps/klipper/files/macros.cfg
Atomique13: not planned ig

meta-opencentauri/recipes-apps/klipper/files/macros.cfg
Atomique13: deleted extra messages

meta-opencentauri/recipes-apps/klipper/files/macros.cfg
Atomique13: removed

meta-opencentauri/recipes-apps/klipper/files/macros.cfg
Atomique13: fixed

meta-opencentauri/recipes-apps/klipper/files/macros.cfg
Atomique13: removed

meta-opencentauri/recipes-apps/klipper/files/macros.cfg
Atomique13: fixed

meta-opencentauri/recipes-apps/klipper/files/macros.cfg
Atomique13: fixed

meta-opencentauri/recipes-apps/klipper/files/macros.cfg
Atomique13: fixed

meta-opencentauri/recipes-apps/klipper/files/macros.cfg
Atomique13: not planned

meta-opencentauri/recipes-apps/klipper/files/macros.cfg
Atomique13: removed

meta-opencentauri/recipes-apps/klipper/files/macros.cfg
Atomique13: moved both pause and cancel in same corner

meta-opencentauri/recipes-apps/klipper/files/macros.cfg
Atomique13: not planned

meta-opencentauri/recipes-apps/klipper/files/macros.cfg
Atomique13: done your majesty

meta-opencentauri/recipes-apps/klipper/files/macros.cfg
Atomique13: get some help

meta-opencentauri/recipes-apps/klipper/files/macros.cfg
Atomique13: removed

meta-opencentauri/recipes-apps/klipper/files/macros.cfg
Atomique13: fixed

meta-opencentauri/recipes-apps/klipper/files/macros.cfg
Atomique13: m83 is for extruder not relative movement

meta-opencentauri/recipes-apps/klipper/files/macros.cfg
Atomique13: ah no it doesn't

meta-opencentauri/recipes-apps/klipper/files/macros.cfg
Atomique13: ah nope, cause users would get bamboozled, this should be a wait until extrusion is done but we don't have that so we let prompt on screen until done

meta-opencentauri/recipes-apps/klipper/files/macros.cfg
Atomique13: if it breaks I am gonna complain

removed

meta-opencentauri/recipes-apps/klipper/files/macros.cfg
Atomique13: removed

meta-opencentauri/recipes-apps/klipper/files/macros.cfg
Atomique13: okey, removed

meta-opencentauri/recipes-apps/klipper/files/macros.cfg
Atomique13: done

meta-opencentauri/recipes-apps/klipper/files/macros.cfg
Atomique13: fixed it

meta-opencentauri/recipes-apps/klipper/files/macros.cfg
Atomique13: have no idea need more research

meta-opencentauri/recipes-apps/klipper/files/macros.cfg
Atomique13: done

meta-opencentauri/recipes-apps/klipper/files/macros.cfg
Atomique13: fixed

meta-opencentauri/recipes-apps/klipper/files/macros.cfg
Atomique13: I removed one last wipe

meta-opencentauri/recipes-apps/klipper/files/macros.cfg
Atomique13: I removed last purge so it does just purge then wipes twice

meta-opencentauri/recipes-apps/klipper/files/macros.cfg
Atomique13: I edited the logic but needs more testing this won't be right

meta-opencentauri/recipes-apps/klipper/files/macros.cfg
Atomique13: removed

meta-opencentauri/recipes-apps/klipper/files/macros.cfg
```diff
@@ -3,252 +3,370 @@
...
+    M190 S{target_bed}
+
+  # rehome after thermal expansion
+    SET_DISPLAY_TEXT MSG="Heatsoak: {heatsoak}m"
+    G4 P{(heatsoak * 60 * 1000)|int}
+    LOADCELL_Z_HOME
+    {% if adaptive_mesh %}
+        M109 S{global.extruder_temp.probing}
+        SET_DISPLAY_TEXT MSG="Bed mesh adaptive"
+        BED_MESH_CALIBRATE ADAPTIVE=1
```
Nopraz: While starting printing ASA I found out that we need to send the bed temp like this in order to reach the correct temp, otherwise it will fall back to 60 :

`BED_MESH_CALIBRATE ADAPTIVE=1 BED_TEMP={target_bed}`

Atomique13: yep

PR #159, opened 2026-04-24
Retry ustreamer http connection in-app

suchmememanyskill:
- Fix /etc/init.d/ustreamer stop
- Remove inbetween shim by making a retry patch for ustreamer


suchmememanyskill:
Tested, works

jamesturton (approved):
Looks good. A much nicer solution that what we had before!

Issue #162, opened 2026-04-29
klipper


uwusempai13576:
just installed latest open centuri and in mainsail i get the following alert:

Dependency: Klipper
The current Klipper version does not support all features of Mainsail. Update Klipper to at least v0.11.0-257.

any ideas why this is happening?

babelbernd-del:
This was already showing up in 0.0.4 beta

suchmememanyskill:
This is not actually an issue. We modify the version klipper reports so you can see the cosmos version in the machine tab in mainsail, but that confuses mainsail as a cosmos version obviously isn't a valid klipper version

jamesturton:
Fixed in #167 !

PR #167, opened 2026-05-04
Fix warning in mainsail for klipper version

jamesturton:
This "fixes" the warning in mainsail for the version of klipper being too old by "breaking" the semver format. This is irrelavent to us as klipper can't be updated seperatly and informing the user of the cosmos version is more effective.


jamesturton:
Fixes #162

jamesturton:
Tested working on my CC

suchmememanyskill (approved):

PR #169, opened 2026-05-04
Advertise hostname in dhcp

### LINKED ISSUE #166: Hostname not given over DHCP
I have multiple CC1s and I would like to be able to identify each one on the network more easily. To do so, I modified `/etc/hostname` and set a unique hostname for each machine.

However, this hostname is not sent over in the DHCP request. My router then uses this to assign a DNS address to the device, so I can refer to that in Orca slicer and other places.

It's not a showstopper. I just made a DHCP reservation and manually specified the hostname for now. Maybe there's a way for me to configure this through SSH, but I would expect it to do this out of the box, hence me making this issue.

jamesturton:
Fixes #166
Also switches to using iproute2 to remove the ip shim that moonraker uses


jamesturton:
Tested working on my machine

suchmememanyskill (approved):

PR #170, opened 2026-05-05
CANVAS support via AFC

suchmememanyskill:
Still a lot to do, but the beginnings are here

Hardware support (CANVAS):
- [X] DRV8833 driver (PWM motor control + hall rotation sensor to keep a (mostly) constant mm/s output)
    - [ ] This implementation is entirely gippitied, should be checked and refactored before release
    - [ ] halt/error pins
- [X] Odometer
- [X] LED
    - [x] PWM LED
- [X] 9V/24V enable
- [ ] RFID
- [ ] Buzzer

Hardware support (Toolhead):
- [X] Filament detection sensor
- [X] Pressure detection (tangle detection) sensor
- [x] Cutter detection
- [x] Cover detection
- [x] New thermistor setup
- [x] remove tangle detection from AFC and instead put it into gcode macros
- [x] Cutter position recovery

COSMOS integration
- [x] Automatic flashing of CANVAS
- [x] Manual flashing of CANVAS
- [x] Build CANVAS klipper firmware
- [x] Build CANVAS katapult firmware
- [x] Build CANVAS official bootloader firmware

AFC integration
- [x] Load
- [x] Unload
- [x] Runout
    - Untested
- [x] Pressure detection
- [x] Eject
    - Eject is disabled
- [X] LED status
- [ ] All settings reimplemented
- [ ] A good default configuration


meta-opencentauri/recipes-apps/klipper/files/0003-drv8833-motor-control.patch
```diff
@@ -0,0 +1,1198 @@
+diff --git a/klippy/extras/drv8833.py b/klippy/extras/drv8833.py
+new file mode 100644
+index 00000000..c6271960
+--- /dev/null
++++ b/klippy/extras/drv8833.py
```
jamesturton, line 5: If you are patching in entire new files it might be cleaner and easier to include the files in the SRC_URI like I did here: https://github.com/OpenCentauri/cosmos/blob/main/meta-opencentauri/recipes-kernel/linux/linux-mainline_%25.bbappend#L9

suchmememanyskill: I'd like to eventually pr this into our fork, but for now a patch is in place as i'm not happy yet with the implementation

meta-opencentauri/recipes-data/klipper-afc-addon/files/canvas.cfg
```diff
@@ -0,0 +1,265 @@
+# -----
```
suchmememanyskill, line 1: TODO: remove tangle detection from AFC and instead put it into gcode macros

Atomique13 (approved):
10/10 🚀

jamesturton (approved):
Colossal work @suchmememanyskill!! This looks ok to merge in my opinion.
Maybe we could include the DRV8833 driver in our kalico repo rather than it living in a patch file, but that's not blocking this from being merged.
Was it intended for the `canvas.cfg` to be included even when canvas option isn't set in `cosmos.conf`?

PR #177, opened 2026-05-09
Check if an oc patched .swu is flashed during emergency flashing

suchmememanyskill:


suchmememanyskill:
Tested, works

meta-opencentauri/recipes-core/usb-automount/files/usb-mount
```diff
@@ -44,11 +44,17 @@ install_firmware_update() {
 
     mv "$SWU_FILE" "$SRC_DIR/emergency.swu.installed" || return 1
 
+    if strings "$SRC_DIR/emergency.swu.installed" | grep -qx "dsp0_partition"; then
+        # Flash stock bootloader
+        restore-mcu-firmware
```
jamesturton: We don't check the return code so we would still try flashing even if something went wrong. Should add `set -e` at the begging of the script or check it manually

jamesturton (approved):
See comment, otherwise looks good

PR #178, opened 2026-05-09
Wipe SAVE_CONFIG from printer.cfg when an update requests it

suchmememanyskill:


suchmememanyskill:
Tested, seems to work

meta-opencentauri/recipes-apps/klipper/files/klipper-init-d
```diff
@@ -15,6 +15,7 @@ PIDFILE="/var/run/klipper.pid"
 KLIPPER_VAR_CONFIG_BUILDER="/usr/bin/build-klipper-var-config"
 KLIPPER_CONFIG_VERSION="1"
 KLIPPER_CONFIG_VERSION_FILE="/etc/config.ver"
```
jamesturton: Could this be `klipper-config.ver` instead?

jamesturton (approved):

PR #185, opened 2026-05-18
Bump grumpyscreen version

jamesturton:
Include our changes as patches and use upstream repo instead


suchmememanyskill (approved):

jamesturton:
Tested working as expected

PR #188, opened 2026-05-20
Klipper: Don't write tmp files to ramdisk

### LINKED ISSUE #87: Bed mcu seems to crash when testing resonnances
How to reproduce:
run `TEST_RESONANCES AXIS=1,1 OUTPUT=raw_data`
observe.

Here are the logs, might be something else entirely making it crash
[klippy(3).log](https://github.com/user-attachments/files/26487392/klippy.3.log)
[moonraker(1).log](https://github.com/user-attachments/files/26487393/moonraker.1.log)

jamesturton:
Hopefully fixes #87. Not tested yet.

### DIFF
--- a/meta-opencentauri/recipes-apps/klipper/files/0001-Don-t-write-tmp-files-to-ramdisk.patch
+++ b/meta-opencentauri/recipes-apps/klipper/files/0001-Don-t-write-tmp-files-to-ramdisk.patch
@@ -1,0 +1,56 @@
+From 5fa49cd66bad8fd2b653c7dc6f820bcd7af10d3a Mon Sep 17 00:00:00 2001
+From: James Turton <james.turton@gmx.com>
+Date: Wed, 20 May 2026 22:38:12 +0200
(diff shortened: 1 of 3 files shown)


suchmememanyskill:
@jamesturton i question if this is the correct fix.

Like, sure, this works technically but
- Can we not just symlink /tmp to /user-resource/.tmp?
- Can we PR to kalico that they start using the tempfile module from python (so we can set a temp directory like via moonraker)?

For now i think i'd prefer the latter

jamesturton:
>     * Can we not just symlink /tmp to /user-resource/.tmp?

Maybe, but this won't work well for any socket files. If it used `/var/log` that would be a better thing to symlink.
 
>     * Can we PR to kalico that they start using the tempfile module from python (so we can set a temp directory like via moonraker)?

I like this too :) Maybe I will do that

suchmememanyskill (approved):

jamesturton:
@suchmememanyskill Does this look any better?

suchmememanyskill:
From a quick glance, looks fine

PR #191, opened 2026-05-20
Test USB mount script restarting wlan0 when a new wifi config is ingested

suchmememanyskill:


jamesturton:
@suchmememanyskill why was this closed? Did it not work? I was testing new grumpyscreen (https://github.com/OpenCentauri/cosmos/pull/303) and it now has a setting for a restart wifi script after connecting to a new ap, so I wondered if this script could be shared between them.

suchmememanyskill:
Never tested it as it wasn't needed; switching from patched to cosmos immediately connected to the right wifi network @jamesturton

PR #193, opened 2026-05-23
Delete logs when going back to stock; check for control type on toolh…

suchmememanyskill:
…ead for calibration


pdscomp (approved):
Cleanup of logs makes sense. Researched the pid variables, this change should allow our PID to not incorrectly show we are calibrated in cases where you are running an alternate control mode like watermark but the extruder_pid_kp happens to be set to something. Appears to be a positive change! Provided it's been tested good with merging!

PR #195, opened 2026-05-29
Add readonly-config-reset command; reset readonly configs on update

suchmememanyskill:


pdscomp (approved):
This all makes sense, and has been tested by Sims. I agree with the logic and the creation of a backup tarball (with correct date ordering!) before cleaning out the overlay.

PR #196, opened 2026-05-29
Implement support zips

suchmememanyskill:


jamesturton (approved):
I haven't tested, but it looks good to me

PR #197, opened 2026-05-29
Bump order of shaper calibration

suchmememanyskill:


pdscomp (approved):
Change only relocates extruder PID and z-offset/mesh after input shaper calibration. Don't see any issues with this! If you've tested the updated calibration macro then I would say we're good to merge!

PR #200, opened 2026-06-06
Don't turn off the screen when stopping the screen service

suchmememanyskill:
instead just blank the screen


Atomique13 (approved):
doesn't look good to me but I still approve, unlike someone else 👀

PR #208, opened 2026-06-14
Add M600 filament-change macro

### LINKED ISSUE #156: Pause and filament change aren’t working
Pause and filament change at a specific layer don’t work. In OrcaSlicer, I tried both pause and filament change at a given layer. Cosmos 0.0.4 with green screen fix.

Image

Image

After the fact, I noticed that Mainsail has an option to set a pause at a specific layer.

Image

 Should I use that instead of the option in Orca?

scv9:
## Summary
Slicers and manual colour swaps emit `M600` to trigger a filament change. No `M600` macro was defined, so Kalico aborted the print with `Unknown command: M600`.

This adds an `M600` macro that:
- pauses the print (parking via the existing `PAUSE` client macro),
- unloads the filament using the existing `UNLOAD_FILAMENT` macro (keeps the current hotend temp so the retract works),
- prompts the operator to load new filament (`LOAD_FILAMENT`) and press `RESUME`.

Fixes #156

## Test plan
- [ ] Start a print, send `M600` mid-print: toolhead parks, filament is cut/ejected, prompt shown.
- [ ] `LOAD_FILAMENT` then `RESUME` continues the print correctly.

### DIFF
--- a/meta-opencentauri/recipes-apps/klipper/files/macros.cfg
+++ b/meta-opencentauri/recipes-apps/klipper/files/macros.cfg
@@ -294,6 +294,23 @@ gcode:
 
     M400                            ; Wait for movements to complete
 
+# Filament change. Slicers (and manual colour swaps) emit M600; without this
+# macro Kalico aborts the print with "Unknown command: M600".
+[gcode_macro M600]
+description: Pause the print, unload filament and wait for a reload
+gcode:
+    {% set x = params.X|default(202)|float %}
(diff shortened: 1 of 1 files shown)


suchmememanyskill:
This is being worked on in #155

scv9:
Thanks — confirmed #155 adds M600 (PAUSE STATE=filament_change + UNLOAD_FILAMENT), which is the more complete approach. Happy to close this as a duplicate.

PR #209, opened 2026-06-14
Guard default bed mesh load on startup

### LINKED ISSUE #110: Suppress `bed_mesh: Unknown profile [default]`
It does that every time I open the WebUI without the default bed mesh.

I have two main bed meshes I use which are named `default on 60` and `default on 80`. So I always wake up to seeing that light error.

scv9:
## Summary
The startup `delayed_gcode` unconditionally ran `BED_MESH_PROFILE LOAD=default`. On a fresh install (and any time the `default` profile is absent) this throws `bed_mesh: Unknown profile [default]` on every boot.

Wrap the load in an existence check using the same condition already used by `_CHECK_CALIBRATION_VARS` in `calibration.cfg`.

Fixes #110

## Test plan
- [ ] Boot with no saved bed mesh: no error, no profile loaded.
- [ ] Create a `default` mesh, reboot: profile loads as before.

### DIFF
--- a/meta-opencentauri/recipes-apps/klipper/files/macros.cfg
+++ b/meta-opencentauri/recipes-apps/klipper/files/macros.cfg
@@ -326,7 +326,12 @@ gcode:
     RUN_SHELL_COMMAND CMD=CAMERA_LIGHT_ON
   {% endif %}
 
-  BED_MESH_PROFILE LOAD=default
+  # Only load the default mesh if it has actually been created. On a fresh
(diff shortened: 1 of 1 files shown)


suchmememanyskill:
The way loading the default mesh works will be changed in #155

scv9:
Thanks — confirmed #155 guards the startup load with `{% if "default" in printer.bed_mesh.profiles %}`, which fixes #110. Superseded, happy to leave this closed.

PR #210, opened 2026-06-14
Turn off heaters after a standalone WIPE_NOZZLE

### LINKED ISSUE #147: WIPE_NOZZLE macro ends with extruder temp on 140 and bed plate 60 even after action is complete
a bit weird to leave the temps hanging.

scv9:
## Summary
`WIPE_NOZZLE` ended by heating the bed to `BED_TEMP` and the nozzle to 140C as preparation for bed meshing. When the macro is run on its own (e.g. a "clean nozzle" action) it left the printer sitting at 140/60 with no print running.

Add a `KEEP_HOT` parameter (default `0`):
- standalone runs now switch the heaters off at the end;
- `BED_MESH_CALIBRATE_WITH_WIPE` passes `KEEP_HOT=1` to preserve the existing calibration behaviour without an extra cool/reheat cycle.

Fixes #147

## Test plan
- [ ] Run `WIPE_NOZZLE` standalone: heaters and fans end at 0.
- [ ] Run `BED_MESH_CALIBRATE_WITH_WIPE`: temps are held through calibration as before.

### DIFF
--- a/meta-opencentauri/recipes-apps/klipper/files/macros.cfg
+++ b/meta-opencentauri/recipes-apps/klipper/files/macros.cfg
@@ -84,6 +84,10 @@ gcode:
 [gcode_macro WIPE_NOZZLE]
 gcode:
     {% set bed_temp = params.BED_TEMP|default(60)|float %}
+    # KEEP_HOT=1 leaves the bed/nozzle at calibration temps for a following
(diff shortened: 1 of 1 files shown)


Eelviny:
#155 is due to be merged soon, a complete rewrite of all macros. Worth checking if the issue was already solved there

suchmememanyskill:
#155 does fix this by restoring the target temp of the hotend after the macro is done

Issue #137, opened 2026-04-13
Home Y before X


seidler2547:
Please change the homing sequence to home Y first, then X. It is very important to home Y first, because the Y-path is always unrestricted, but the X path has obstructions like the tray lever and the stepper motor in the back. Homing X first is very dangerous.

suchmememanyskill:
Y is not unrestricted. If X is left at or near max, you will engage the filament cutter if you home Y first. 

There is no good way to home the CC. It's a mess.

seidler2547:
But it's still more reasonable, and also how stock firmware does it. I think stock even moves a little left before homing Y so that it would not run into the filament cutter.

suchmememanyskill:
Stock doesn't do it like this. We are currently identical to stock, home x, home y, home x, home y. Stock homes x left though, not right.

PR #212, opened 2026-06-14
Home Y before X in homing_override

scv9:
## Summary
The `homing_override` homed X first, which (per the existing in-code comment) can drive the toolhead into the rear steppers/frame depending on the start position. This reorders the sensorless homing sequence to **Y, X, Y, X**.

Refs #137

> [!WARNING]
> Motion-order change. I do not have hardware to validate this end-to-end — please verify sensorless homing on a real ECC1 before merging.

## Test plan
- [ ] `G28` from several start positions: no collision with rear frame, X/Y home reliably.

### DIFF
--- a/meta-opencentauri/recipes-apps/klipper/files/macros.cfg
+++ b/meta-opencentauri/recipes-apps/klipper/files/macros.cfg
@@ -369,24 +369,26 @@ gcode:
       M400
     {% endif %}
 
-    # First X home. Note this may crash into the back steppers.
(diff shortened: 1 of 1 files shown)


suchmememanyskill:
This is being worked on in #155

scv9:
Thanks — confirmed #155 reworks homing_override to `axes: yxz` (Y before X), which addresses #137. Superseded, happy to leave this closed.

PR #213, opened 2026-06-14
Explicitly enable force_move

scv9:
## Summary
`macros.cfg` uses `FORCE_MOVE` / `SET_KINEMATIC_POSITION` (homing_override Z clearance when Z is unhomed, `STRAINGAUGE_Z_HOME`, `UNLOAD_FILAMENT`), but no `[force_move]` section enabled them. In stock Klipper these commands require `enable_force_move: True`. Add it so homing/unload do not depend on Kalico defaults.

> [!NOTE]
> Hardening change. Please confirm on hardware that homing/unload still behave as expected.

## Test plan
- [ ] `G28` with Z unhomed: Z-clearance force move works.
- [ ] `UNLOAD_FILAMENT` / `STRAINGAUGE_Z_HOME` work as before.

### DIFF
--- a/meta-opencentauri/recipes-apps/klipper/files/machine.cfg
+++ b/meta-opencentauri/recipes-apps/klipper/files/machine.cfg
@@ -20,6 +20,12 @@ restart_method: command
 [danger_options]
 multi_mcu_trsync_timeout: 0.03
 
+# macros.cfg relies on FORCE_MOVE / SET_KINEMATIC_POSITION (homing_override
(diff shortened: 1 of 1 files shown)


jamesturton:
What problem are you trying to solve here? This is already enabled by default in kalico and we do not support standard klipper in cosmos. We could go through and explicitly define all default values in our config but that isn't really a good idea as that will only make it even more difficult to read.

scv9:
Thanks — you're right, force_move is enabled by default in Kalico and COSMOS doesn't target stock Klipper, so this is unnecessary. Closing.

PR #215, opened 2026-06-19
Preserve default config comments in rewritten configs

suchmememanyskill:
## Summary
- Read comments from `default.conf` and carry them into rewritten config output
- Render config files manually so existing inline documentation is preserved
- Skip unnecessary writes when the rendered file already matches disk content

## Testing
- Not run (not requested)


suchmememanyskill:
Needs some more testing but seems to work

PR #216, opened 2026-06-20
Remove unused park variables from Klipper macros

suchmememanyskill:
## Summary
- Replace the unused pause and cancel park dictionaries with a single retract amount
- Update `PAUSE` and `KICK` to use the shared retract value
- Keep the existing retract behavior while simplifying the macro configuration

## Testing
- Not run (not requested)


suchmememanyskill:
Tested, works

PR #221, opened 2026-06-21
Fix target in _LOAD_FILAMENT_STEP_PUSH

suchmememanyskill:


meta-opencentauri/recipes-apps/klipper/files/macros.cfg
```diff
@@ -348,6 +348,9 @@ gcode:
 
 [gcode_macro _LOAD_FILAMENT_STEP_PUSH]
 gcode:
+    {% set global = printer['gcode_macro _global_var'] %}
+    {% set target = params.EXTRUDER_TEMP | default(global.extruder_temp.loading) | int %}
```
suchmememanyskill: This should probably just be global.extruder_temp.loading

suchmememanyskill:
maybe lets error instead?

PR #223, opened 2026-06-22
Force MZV shaper in FULL_CALIBRATION

suchmememanyskill:


pdscomp:
Reviewed change, if this shaper will provide good results across a range of CC1 printers than it seems like worth doing!

PR #232, opened 2026-06-30
Rotate klipper and moonraker logs at 5mb

jamesturton:
Use RotatingFileHandler rather than TimedRotatingFileHandler to rotate when the logs get to 100kb rather than every hour.


Atomique13 (commented):
klippy.log gets rotated at 5mb but not moonraker.log, moonraker.log got rotated after firmware restart tho.

PS: Works, moonraker gets rotated when it writes smth to it, not when I do it manually. So it works, it got rotated once a proper log continued.

PPS: Gets rotated and also deletes last 4th backup. 

Image

Atomique13 (approved):

Issue #236, opened 2026-07-05
Add support for skew_profile in PRINT_START macro


bniggemyer:
If the user has created a skew profile, enable loading the skew profile in the PRINT_START macro.  Per the documentation, loading of skew profile should occur after any homing or other non-print toolhead movement.  This is to reduce the small chance of sending the toolhead outside the boundaries.   Because PRINT_START can move to tray and also home, the loading of skew profile needs to occur after those activities.

https://www.klipper3d.org/Skew_Correction.html

PR #237, opened 2026-07-05
Add support for skew_profile

bniggemyer:
https://github.com/OpenCentauri/cosmos/issues/236
Added support to load skew profile in PRINT_START and clear profile in PRINT_END

Test1 :  
-set load_skew_profile = True in cosmos.conf
-initiated test print
-ran GET_CURRENT_SKEW during print to confirm the profile was loaded
-ran GET_CURRENT_SKEW following the print to confirm the profile was cleared

Test2:
-set load_skew_profile = False in cosmos.conf
-initiated test print
-ran GET_CURRENT_SKEW during print to confirm no skew was loaded


bniggemyer:
discussed with Sims, closing current PR, will resubmit simplified version

PR #242, opened 2026-07-08
Implement update check

suchmememanyskill:


suchmememanyskill:
image

image

meta-opencentauri/recipes-data/config-manager/files/__pycache__/config_manager.cpython-314.pyc
jamesturton, line 1: I assume it was not intentional to commit this bytecode?

jamesturton (changes requested):
Looks good to me other than the python bytecode file which seems to have snuck in

jamesturton (approved):

PR #248, opened 2026-07-15
added manual level and check bed level macros

Bad-Hackerman:
Manual_level macro which uses stc to level the bed to be flat and at the right height for probing

Integration in calibration script not done as it is getting reworked.

Changed screw pitch in config from M3 to M4


pdscomp:
Frontend counterpart is up: #297 flips Fluidd's and Mainsail's built-in Screws Tilt Adjust dialogs to off-by-default in the shipped image, so this PR's prompt-based leveling flow is the only thing on screen. Re-enable is one settings toggle in either frontend. Best merged together—without this one, stock STC users get console-text-only output.

Atomique13 (approved):

PR #250, opened 2026-07-15
printing and loading turns on toolhead led

Bad-Hackerman:
printing and loading turns on toolhead led, and turns them off once done


suchmememanyskill:
@Bad-Hackerman Can this be an option in cosmos.conf?

Bad-Hackerman:
oh and added a load cancel marco in order to turn hotend an led off

suchmememanyskill (approved):

PR #252, opened 2026-07-19
Reduce CC1 runtime memory pressure and make calibration/update paths more reliable

pdscomp:
# Reduce CC1 runtime memory pressure and make calibration/update paths more reliable

## Summary

Centauri Carbon 1 has 128 MiB of RAM shared by Klipper, Moonraker, the UI, camera streaming, and calibration.  This series reduces steady-state memory retention, bounds the largest calibration allocations, adds a lower-RAM calibration implementation, and makes two reliability-sensitive boot/update paths deterministic:

- uStreamer uses substantially less resident memory while preserving snapshots and live video;
- Klipper and Moonraker start with lower Python and allocator overhead;
- Kalico no longer materialises input-shaper FFT intermediates for every capture window at once;
- rusty-shaper produces the same result as Classic calibration in markedly less time and with considerably more memory and swap headroom; and
- zram startup and image-owned SysV init-script refreshes are observable, bounded, and robust across updates.

The changes are deliberately split into focused commits so the kernel policy, Python runtime policy, Kalico algorithm change, and optional rusty-shaper workflow can be reviewed or reverted independently.

## Upstream dependency

This PR depends on [OpenCentauri/OpenCentauri#108](https://github.com/OpenCentauri/OpenCentauri/pull/108), which adds the GPL-3.0 rusty-shaper implementation, regression fixtures, verification gate, and documented integration examples. Merge #108 first; this Cosmos PR packages and integrates that application into the CC1 image.

## Commit series

| Commit | Change | Why it matters on CC1 |
| --- | --- | --- |
| `b926d5d` | `zram: make swap initialization reliable on low-memory printers` | A boot without zram is much more likely to OOM during normal service load or calibration. |
| `764e174` | `overlayfs: refresh image-owned init scripts on update` | A persistent `/etc` overlay previously could keep stale service launch scripts after an image update, silently defeating memory/runtime fixes. |
| `4831b0a` | `ustreamer: reduce camera streamer memory use` | Reduces always-on camera memory use without sacrificing the normal snapshot/live-view workflow. |
| `78ad740` | `overlayfs: escape shell expansions in preinit template` | Fixes rendering of the image-owned init-script cleanup in the pre-init template, allowing the update-path fix to build and execute correctly. |
| `b5d80fc` | `kernel: reduce runtime memory footprint on Centauri` | Removes unused kernel facilities and their runtime state from the appliance profile. |
| `4edb6be` | `python: reduce Klipper and Moonraker memory overhead` | Reduces Python code-object/cache and glibc allocator retention in the two largest long-lived user-space services. |
| `c25ea7b` | `kalico: bound input-shaper calibration memory` | Keeps Classic input-shaper processing from allocating FFT/window temporaries proportional to the complete accelerometer capture. |
| `c851dcb` | `rusty-shaper: add low-RAM input-shaper calibration` | Adds the validated rusty-shaper implementation as the default selectable, non-disruptive calibration path while retaining Classic as a fallback. |
| `51cda0e` | `calibration: skip stages with saved results` | Avoids unnecessary motion, heating, camera/UI disruption, and calibration-time memory pressure while retaining a deliberate full-recalibration path. |
| `6f900c9` | `kalico: persist input-shaper settings and stream raw accelerometer writes` | `SET_INPUT_SHAPER` now marks `[input_shaper]` dirty so `SAVE_CONFIG` persists rusty-shaper/Classic results; raw accelerometer captures are streamed to CSV instead of building a full samples list. |
| `b348004` | `calibration: detect only saved extruder PID results` | Prevents factory/default PID coefficients from being reported as user-calibrated. |

### Reliable zram and update behaviour

`b926d5d` replaces the previous one-shot zram setup with a bounded, idempotent initializer.  It waits for the device node/sysfs attributes, chooses the first available compressor, verifies that `/dev/zram0` is actually active in `/proc/swaps`, and retries transient failures once through a module reload.  It logs state and failures rather than silently continuing without swap.  The existing CC1 tuning remains (`swappiness=150`, 200% logical zram size, priority 100), but failure to bring swap online is now diagnosable.

`764e174` treats package-owned `/etc/init.d` scripts as release-owned before mounting the persistent `/etc` overlay.  On the first boot after an A/B update it removes only upper-layer copies or whiteouts that shadow a script present in the new image.  User-created scripts remain intact.  This is important for the Moonraker, uStreamer, and zram changes in this branch: old overlay copies can no longer leave a newly flashed image running an old service command. `78ad740` correctly escapes shell parameter expansions in the Python-rendered pre-init template, so this cleanup is build-safe and renders to the intended shell script.

### uStreamer: lower camera memory with live video retained

`4831b0a` keeps the production camera command at 864x480, camera-provided MJPEG, one worker/buffer, and 8 FPS:
```text
/usr/bin/ustreamer -r 864x480 -c HW -m MJPEG -b 1 -w 1 -f 8 -l \
  --host=0.0.0.0 --allow-origin=*
```

The patch caps shallow worker stacks, keeps a single newest HTTP JPEG rather than retaining stale frame copies, safely borrows camera JPEG data for HTTP delivery, drops the permanent blank RGB canvas, and starts generic frame buffers small enough to grow on demand.  It tries bounded USERPTR capture when supported and retries with MMAP when a UVC device rejects USERPTR; the deployed CC1 UVC camera was validated using the MMAP fallback.

With an active 864x480/8 FPS client on Centauri, uStreamer PSS fell from **4.67 MiB to 3.08 MiB** (1.59 MiB / 34% less) and RSS fell from **5.76 MiB to
4.14 MiB** (1.62 MiB / 28% less).  Snapshots and the live stream were verified
after a CC1 eMMC build and flash.

### Kernel and Python baseline reductions

`b5d80fc` adds the final CC1 kernel memory profile.  It disables unused debug and symbol tables, tracing/perf/task accounting, cgroups/container features, io_uring and other unused general-purpose interfaces, IPv6/firewall/bridge/ VLAN networking, and non-CC1 Allwinner pinctrl/PWM drivers.  It explicitly retains the interfaces required by the appliance: display/DRM, UVC media, RPMsg/remoteproc, GPIO/I2C/SPI, USB storage, swap, and zram.  The intent is less resident kernel code/state and fewer background accounting structures; this configuration change is not represented as a separate live MiB claim.

`4edb6be` starts Klipper and Moonraker with `-O -X no_debug_ranges`, ships target-version optimized bytecode without build-directory paths/debug ranges, and sets `MALLOC_ARENA_MAX=1` in both service environments.  The first two reduce Python code/cache overhead; the arena cap prevents glibc from retaining multiple per-thread heaps on a 128 MiB machine.  An isolated live Centauri test attributed approximately **3 MiB of Klipper PSS** reduction to the arena limit while the printer API remained healthy.  The change does not remove any Moonraker components or alter its public API.

### Bound Classic Kalico calibration allocations

`c25ea7b` changes Kalico's Welch PSD calculation to accumulate FFT energy in fixed 64-window batches rather than creating detrended/windowed/FFT arrays for all overlapping windows simultaneously.  It also retains accelerometer data as `float32` and uses timestamps relative to the first sample before packing. The raw capture CSV contract and PSD calculation are preserved, while the largest temporary arrays are bounded by batch size instead of capture length. This makes Classic calibration safer under memory pressure even when rusty-shaper is not selected.

### rusty-shaper: faster, lower-pressure calibration with equal output

`c851dcb` packages the stripped GPL-3 rusty-shaper binary and its fully checksummed Cargo dependencies, then integrates it into the image, settings, and calibration macros.  rusty-shaper is selectable through `input_shaper` (default `rusty`); Classic remains selectable as the fallback. Candidate shapers are configurable. The selective `FORCE=0` path skips load-cell tare work when a saved tare is present; the default full calibration deliberately reruns it.

The G-code flow captures X and Y resonance data, then schedules a detached worker after the capture macro has returned.  This is critical: a synchronous `RUN_SHELL_COMMAND` blocks Klipper's reactor, while rusty-shaper reports progress and completion through Moonraker/M125.  The worker therefore parses the most recent X/Y raw captures, runs rusty-shaper, writes a success/error status file, emits Moonraker-visible status messages, and requests the remaining full-calibration steps only when it was invoked as part of full calibration.  A standalone rusty-shaper run applies its shaper setting and exits; it does not unexpectedly run PID or mesh calibration.

`51cda0e` makes the calibration entry point saved-state-aware. `CALIBRATE_ALL` defaults to `FORCE=1`, preserving the established full recalibration behavior. In contrast, `CHECK_CALIBRATION` launches `CALIBRATE_ALL FORCE=0`; that selective path runs only missing input-shaper, load-cell tare, extruder PID, and default-mesh stages. `CALIBRATE_ALL FORCE=0` can also be used directly to request that behavior.

The macros record the stages actually run so that `SAVE_CONFIG` is still offered whenever a new shaper, tare, PID, or mesh result needs persistence. They use the explicit saved `configfile.config` extruder `pid_version` marker, rather than Kalico's synthesized runtime `configfile.settings` value or a magic PID coefficient. `b348004` prevents factory/default PID coefficients from being reported as user-calibrated. The flow also removes its redundant standalone `CLEAN_NOZZLE`: the existing `BED_MESH_CALIBRATE` wrapper already cleans the nozzle when a mesh is needed. Prompt text documents the default and `FORCE=1` behavior and only asks the operator to clear the chamber/bed when a remaining stage requires it.

#### Kalico-parity regression coverage

rusty-shaper has deterministic real-capture regression coverage in its normal host verification gate, `rusty-shaper/check.sh` (`cargo test --locked`, strict Clippy, and a locked release build).  `cargo test --locked` includes `scorer::tests::real_world_regression::{all_captures_match_kalico_recommendation,all_captures_match_kalico_per_shaper_metrics}`.  Those tests replay fourteen committed, Git-LFS-compressed raw X/Y captures from seven attributed community datasets.  Their JSON goldens are generated by Kalico's `scripts/calibrate_shaper.py` (currently recorded against Kalico `693cd75b`), rather than by rusty-shaper itself.

For every capture, the first test requires the exact same recommended shaper name and a frequency within 0.5 Hz of Kalico.  The second also checks every candidate's fitted frequency, residual vibration, smoothing, and `max_accel` against explicit tolerances.  Both tests use rusty-shaper's production default shaper set and the same normalize/low-frequency-suppression pipeline as the CLI, so they exercise the deployed calibration path rather than synthetic unit data.

The current Yocto `rusty-shaper` recipe compiles the component but does not invoke this host test gate, and the upstream worktree does not yet have a rusty-shaper-specific GitHub Actions workflow.  Thus the suite is the maintained component verification gate, but it is not honestly describable as running on every Cosmos image build today; wiring `./check.sh` into CI is a recommended follow-up before making that stronger claim.

#### CC1 benchmark

Both standalone paths were run on the same flashed CC1 while collecting memory/swap/process telemetry capped at 2 Hz (the serial SSH collection loop achieved approximately 0.5 Hz). Each completed successfully and selected exactly the same shaper:
```text
Rusty:   MZV, X 55.4 Hz, Y 47.2 Hz
Classic: MZV, X 55.4 Hz, Y 47.2 Hz
```

| Metric | Rusty | Classic | Rusty advantage |
| --- | ---: | ---: | ---: |
| End-to-end calibration time | 70.2 s | 221.2 s | 151.0 s faster (68.3%) |
| Peak used RAM | 76.58% | 86.67% | 10.09 percentage points lower |
| Minimum `MemAvailable` | 26.82 MiB | 15.26 MiB | 11.56 MiB more headroom |
| Peak swap used | 21.25 MiB | 55.75 MiB | 34.50 MiB less swap pressure |
| Time with RAM use >=75% | 8.3 s | 79.8 s | 71.5 s less high-pressure time |
| Time with RAM use >=80% | 0 s | 8.8 s | avoids the severe-pressure interval |

Classic deliberately stops Grumpy and uStreamer around its calibration work; Rusty leaves both services running.  Consequently the global RAM/swap result is conservative: Rusty achieved the lower peak and shorter high-pressure period while carrying the normal UI and live-camera workload that Classic had temporarily removed.  Klipper/Moonraker remained healthy after both runs, heater targets were zero, and neither standalone run wrote the calibration result to `printer.cfg`.

#### Idle memory improvement

A side-by-side idle measurement on the same CC1 flashed with the pre- optimization baseline versus this branch shows a steady-state gain:

| Metric | Pre-optimization | This branch | Improvement |
| --- | ---: | ---: | ---: |
| `MemAvailable` | 29.60 MiB | 35.64 MiB | 6.04 MiB |
| Used RAM | 76.14 MiB | 70.82 MiB | 5.32 MiB |
| Swap used | 10.75 MiB | 10.00 MiB | 0.75 MiB |

The improvement is dominated by the Python arena cap, kernel footprint reduction, and lower steady-state allocator retention from the uStreamer and Kalico changes.  Swap stays essentially flat because both images already had reliable zram online.

## Validation

- Built CC1 eMMC image and flashed a CC1 for the uStreamer deployment test; snapshot and active live video remained functional at 864x480/8 FPS.
- Validated the UVC camera's USERPTR rejection and MMAP retry path rather than treating an unsuccessful USERPTR capture as a memory result.
- Verified the rusty-shaper and Classic standalone calibration macros on the flashed CC1, including the rusty-shaper asynchronous completion path, Moonraker/M125 messages, raw X/Y capture discovery, results, and idle recovery.
- Collected the Rusty-vs-Classic comparison from raw time-series telemetry; the raw captures and analysis are retained with the test artifacts.
- Kept the full-calibration continuation gated on an explicit `FULL=1` worker completion, preventing standalone calibration from entering PID/mesh steps.

## Operational notes

- The rusty-shaper package is GPL-3.0-only because it is derived in part from Klipper/Kalico behavior; its recipe declares that license and ships the corresponding license file checksum.
- This PR intentionally preserves Classic calibration as an operational fallback and makes its own memory behavior safer through the Kalico batch patch.
- The kernel feature reductions are scoped to the CC1 appliance profile; required hardware and zram options are explicitly pinned on.


jamesturton (changes requested):
Due to the size of this PR I think it's better that it is split up this up into smaller PR, one for each feature/fix made, in order to help speed up being able to review and test each change made. Certain changes such as the new shaper calibration module should probably be tested by a small focus group to ensure that is had the same result as the original module.

meta-opencentauri/recipes-apps/klipper/kalico_2026.02.00.bb
```diff
@@ -82,9 +82,10 @@ do_install() {
     sed -i 's/APP_NAME = "Kalico"/APP_NAME = "${DISTRO_NAME}"/' ${D}${datadir}/klipper/klippy/__init__.py
     echo "Release - ${DISTRO_VERSION}" > ${D}${datadir}/klipper/klippy/.version
 
-    # Remove any .pyc files to avoid TMPDIR references
+    # Remove any source-tree caches, then ship target-version optimized
+    # bytecode without debug ranges.  Stripping ${D} avoids TMPDIR references.
     find ${D} -name '*.pyc' -delete
-    find ${D} -name '__pycache__' -type d -exec rm -rf {} + 2>/dev/null || true
+    PYTHONNODEBUGRANGES=1 ${PYTHON} -O -m compileall -q -s ${D} ${D}${datadir}/klipper
```
jamesturton, line 88: Compile should be in `do_compile` not `do_install`

meta-opencentauri/recipes-apps/moonraker/moonraker_0.10.0.bb
```diff
@@ -70,6 +70,10 @@ do_install() {
     # Install moonraker python package
     install -d ${D}${datadir}/moonraker
     cp -r ${S}/moonraker ${D}${datadir}/moonraker/
+    # Ship target-version optimized bytecode.  Removing debug ranges makes
+    # code objects and their cache files smaller; -s keeps build paths out of
+    # tracebacks and avoids a TMPDIR-dependent cache.
+    PYTHONNODEBUGRANGES=1 ${PYTHON} -O -m compileall -q -s ${D} ${D}${datadir}/moonraker
```
jamesturton, line 76: Compile should be in `do_compile` not `do_install`

pdscomp:
## Meta feedback from @suchmememanyskill

- PR too big; each patch's contribution needs to be understood separately (in progress — splitting into per-topic PRs with per-PR numbers).
- Patches look not very targeted; the goal per patch was unclear.
- Caution against retroactively updating the implementation mid-discussion.

meta-opencentauri/recipes-apps/klipper/kalico_2026.02.00.bb
pdscomp: With python wouldn't it be the other way around though? You would deploy the .py files to the staging sir, then run the python compiler to generate pyc? I don't think it would work the other way, you need the py in their final resting place first?

meta-opencentauri/recipes-apps/moonraker/moonraker_0.10.0.bb
pdscomp: With python wouldn't it be the other way around though? You would deploy the .py files to the staging sir, then run the python compiler to generate pyc? I don't think it would work the other way, you need the py in their final resting place first?

PR #253, opened 2026-07-19
fix: use target headers when building atomscreen

mschfh:
This PR updates the `atomscreen` recipe to use the target sysroot headers instead of the host ones:
```
| error: failed to run custom build command for `linuxfb v0.3.1`
|   Unable to generate bindings: ClangDiagnostic("/usr/include/linux/types.h:5:10: fatal error: 'asm/types.h' file not found\n")
```


jamesturton:
This is basically a dupe of #79, which I have now merged. Hopefully this fixes your issue too.

PR #254, opened 2026-07-19
zram/overlayfs: reliable swap init and image-owned init-script refresh

pdscomp:
# zram/overlayfs: reliable swap init and image-owned init-script refresh

Split from #252.

## Why this PR

This revision greatly simplifies zram initialization compared with the original version of PR #252. It more closely follows the original script from @jamesturton while retaining added logging, comprehensive state and device checks, idempotent behavior, and bounded retry logic (as well as a test case to validate script execution on the host after any updates).

A CC1 boot that fails to bring zram online is the main OOM path during normal load or calibration. The old one-shot zram init could silently continue without swap. Separately, the persistent `/etc` overlay could keep stale package-owned init scripts after an image update, silently defeating runtime fixes.

## Justification / measurements

Reliability change, not a performance change — no MiB claim.

- New zram init: idempotent, waits for device/sysfs, picks first available compressor, verifies `/dev/zram0` in `/proc/swaps`, retries once via module reload, logs failures.
- Overlay refresh: on first boot after A/B update, removes only upper-layer copies/whiteouts shadowing scripts present in the new image. User-created scripts untouched.

## Risk / rollback

Low. Reverts cleanly; worst case is reverting to previous one-shot zram init behavior.

### DIFF
--- a/meta-opencentauri/images/files/overlayfs-etc-preinit.sh.in
+++ b/meta-opencentauri/images/files/overlayfs-etc-preinit.sh.in
@@ -58,6 +58,34 @@ then
     mkdir -p $UPPER_DIR
     mkdir -p $WORK_DIR
 
+    # Remove local overlay edits to Cosmos-provided init scripts so upgrades
+    # use the new image versions. User-provided init scripts are preserved,
+    # matching previous Cosmos behavior; if an update breaks one, it is the
+    # user's responsibility to fix or remove it.
+    if [ -d "$UPPER_DIR/init.d" ]; then
+        for upper_script in "$UPPER_DIR"/init.d/*; do
+            [ -f "$upper_script" ] || continue
+            script_name=${{upper_script##*/}}
+            [ -f "/etc/init.d/$script_name" ] || continue
+            rm -f "$upper_script"
+        done
+
+        # Remove whiteouts only when they hide an image-owned init script.
+        for whiteout in "$UPPER_DIR"/init.d/.wh.*; do
+            [ -e "$whiteout" ] || continue
+            whiteout_name=${{whiteout##*/}}
+            case "$whiteout_name" in
+                .wh..wh..opq)
+                    rm -f "$whiteout"
+                    ;;
+                .wh.*)
+                    script_name=${{whiteout_name#.wh.}}
+                    [ -f "/etc/init.d/$script_name" ] && rm -f "$whiteout"
+                    ;;
+            esac
+        done
+    fi
+
     if {OVERLAYFS_ETC_EXPOSE_LOWER}; then
         mkdir -p $LOWER_DIR
 
--- a/meta-opencentauri/recipes-core/zram/files/init
+++ b/meta-opencentauri/recipes-core/zram/files/init
@@ -1,72 +1,84 @@
 #!/bin/sh
 ### BEGIN INIT INFO
-# Provides: zram
+# Provides:          zram
 # Required-Start:
 # Required-Stop:
 # Default-Start: 2 3 4 5
 # Default-Stop: 0 1 6
-# Short-Description: Increased Performance In Linux With zRam (Virtual Swap Compressed in RAM)
-# Description: Adapted from zram init script in meta-openembedded
+# Short-Description: zram swap with retries and logging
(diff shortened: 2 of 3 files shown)


pdscomp:

- `overlayfs-etc-preinit.sh.in`: doesn't know what it's supposed to do.
- init.d cleanup "should just be a single line `rm -f /data/overlay-etc/upper/init.d/*`" — user services should be re-symlinked manually per update, since the update could have broken their service anyway.

pdscomp:
@suchmememanyskill

1. The overlayfs preinit block removes overlay copies of image-owned init scripts before /etc is mounted, so an OTA update's new init script isn't shadowed by a stale overlay copy. It only deletes files that also exist in the image's /etc/init.d.

2. A single rm -f /data/overlay-etc/upper/init.d/* would delete user-added init scripts too. The current loop only removes overlays of image-owned scripts, matching the existing behavior we already had for other services.

pdscomp:
## Validation

Ran `meta-opencentauri/recipes-core/zram/tests/init-test.sh` against the new zram init script.

All cases passed:

- **Normal startup:** PASS — zram swap activates successfully.
- **Startup logging:** PASS — successful activation is logged.
- **Idempotent startup:** PASS — a second `start` detects active swap and is a no-op.
- **Transient module failure:** PASS — recovers after one `modprobe` failure and succeeds on retry.
- **Retry logging:** PASS — retry attempt is recorded.
- **Permanent module failure:** PASS — setup fails cleanly after the bounded five-attempt limit.
- **Failure cleanup:** PASS — no swap remains active after permanent failure.

Result: `PASS`

On-machine testing will be next once a formal PR build is complete.

pdscomp:
## On-machine CC1 validation

Validated the formal PR-254 build on `carbon2u` after reboot.

### zram/swap

- `/proc/swaps`: PASS — `/dev/zram0` active, 229,308 KiB total, 9,216 KiB used, priority 100.
- `swapon --show`: PASS — approximately 223.9 MiB zram swap online.
- `free -m`: PASS — approximately 28.4 MiB RAM available with zram enabled.
- Boot log: PASS — kernel added `zram0`, configured capacity, and enabled swap successfully.
- Deployed init script: PASS — contains retry, verification, idempotence, and failure-bounding logic.
- Live idempotence check: PASS — running `/etc/init.d/zram start` while active returned `zram: already active`.
- VM tuning: PASS — swappiness 150, watermark boost 0, watermark scale 125, page-cluster 0, and vfs cache pressure 500.
- No zram or out-of-memory failures were found in the inspected logs.

### Service health

- Klipper: PASS — running and API state `ready`.
- Moonraker: PASS — connected to Klipper, no failed components reported.
- uStreamer: PASS — running.
- Printer API: PASS — reports `Release - PR-254-9b0`, state `ready`.
- System uptime/load: PASS — approximately 7 minutes uptime, load average below 0.1.

The PR is now ready for review and possible merge.

kyleinoregon:
Traced this on a CC1 in #312: the boot failure is `echo 1 > /sys/block/zram0/reset` returning EBUSY because udev still has the freshly created device open, and with `set -e` the script dies there before swap is enabled. #312 keeps the existing script and just waits for `udevadm settle` and retries the reset. If the rewrite here is preferred, it would be worth adding that settle before the reset too, since the 1 s sleep between attempts is what makes the retry succeed today.

pdscomp:
> Traced this on a CC1 in #312: the boot failure is `echo 1 > /sys/block/zram0/reset` returning EBUSY because udev still has the freshly created device open, and with `set -e` the script dies there before swap is enabled. #312 keeps the existing script and just waits for `udevadm settle` and retries the reset. If the rewrite here is preferred, it would be worth adding that settle before the reset too, since the 1 s sleep between attempts is what makes the retry succeed today.

Thanks for the feedback! I updated the script with the udev settle, I think that makes a lot of sense! 
33fda43 zram: settle udev before resetting the device

Atomique13 (commented):
works on my machine, printing too. check discord for more info

@jamesturton can you pretty please check if alright to merge? Thank you 🤗

pdscomp:

PR #255, opened 2026-07-19
ustreamer: reduce camera memory by ~1.6 MiB PSS at 864x480/8 FPS

pdscomp:
# ustreamer: reduce camera memory by ~1.6 MiB PSS at 864x480

Split from #252.

## Why this PR

uStreamer is always on. On a 128 MiB machine, cutting its resident set by a third frees real headroom for calibration and printing without losing the snapshot/live-view workflow.

## Justification / measurements (isolated, live CC1, active 864x480 client)

| Metric | Before | After | Delta |
|---|---:|---:|---:|
| PSS | 4.67 MiB | 3.08 MiB | −1.59 MiB (34%) |
| RSS | 5.76 MiB | 4.14 MiB | −1.62 MiB (28%) |

- Verified on flashed CC1: snapshots and live video functional.
- Production command unchanged in shape: 864x480, camera MJPEG, 1 worker/buffer.

## Commit

- `1d791e3` ustreamer: reduce camera streamer memory use

## Risk / rollback

Low. Config revert restores previous command line; USERPTR attempt safely falls back to MMAP on cameras that reject it (validated on the CC1 UVC camera).

### DIFF
--- a/meta-opencentauri/recipes-apps/ustreamer/files/0002-reduce-memory-usage.patch
+++ b/meta-opencentauri/recipes-apps/ustreamer/files/0002-reduce-memory-usage.patch
@@ -1,0 +1,371 @@
+From 144318c470cb5512227f24d217a7539de47be201 Mon Sep 17 00:00:00 2001
+From: Paul Swenson <pdscomp@gmail.com>
+Date: Fri, 10 Jul 2026 10:15:31 -0400
+Subject: [PATCH] ustreamer: reduce camera streamer memory use
+
+The Centauri Carbon camera service shares a 128 MiB system with the
+printer-control stack, so its steady-state frame buffering matters.  At
+864x480 camera-MJPEG and 8 FPS on carbon2u, the stock ustreamer measured
+4.67 MiB PSS and 5.76 MiB RSS.  This change measured 3.17 MiB PSS and
+4.30 MiB RSS while snapshots and a live stream remained functional.
+
+Reduce long-lived allocations without changing the service architecture:
+
+* cap internal thread stacks;
+* retain only the newest JPEG in the streamer backlog and let HTTP clients
+  borrow a frame while it remains valid;
+* avoid allocating a permanent RGB canvas for the blank frame;
+* start generic frame buffers smaller and grow them on demand; and
+* use bounded USERPTR buffers when supported, with an MMAP retry when a UVC
+  camera rejects USERPTR.
+
+The CC2 UVC camera rejects USERPTR queueing, so MMAP remains the validated
+capture method on this device.  The fallback keeps the safer low-allocation
+default for cameras that do support USERPTR without sacrificing operation on
+the deployed camera.
+
+Signed-off-by: Paul Swenson <pdscomp@gmail.com>
+Upstream-Status: Pending [Not submitted to upstream yet]
+---
+ src/libs/capture.c                  | 37 ++++++++++++++++++++++++++++---
+ src/libs/frame.c                    | 44 +++++++++++++++++++++++++++++++++++--
+ src/libs/frame.h                    |  3 +++
+ src/libs/frametext.c                |  1 +
+ src/libs/queue.c                    |  2 +-
+ src/libs/threading.h                | 36 +++++++++++++++++++++++++++++-
+ src/ustreamer/blank.c               |  1 +
+ src/ustreamer/encoders/hw/encoder.c |  4 +++-
+ src/ustreamer/options.c             |  2 +-
+ src/ustreamer/stream.c              | 13 ++++++++---
+ 10 files changed, 131 insertions(+), 12 deletions(-)
+
+diff --git a/src/libs/capture.c b/src/libs/capture.c
+index ee2021e17dd23083d45dd7a37678efff1adcc817..2614e117120d1a9fdb1c109c62758d526873cd71 100644
+--- a/src/libs/capture.c
++++ b/src/libs/capture.c
+@@ -94,6 +94,8 @@ static int _capture_open_dv_timings(us_capture_s *cap, bool apply);
+ static int _capture_open_format(us_capture_s *cap, bool first);
+ static void _capture_open_hw_fps(us_capture_s *cap);
+ static void _capture_open_jpeg_quality(us_capture_s *cap);
++static uz _capture_get_userptr_buffer_size(const us_capture_runtime_s *run);
++static int _capture_open_impl(us_capture_s *cap);
+ static int _capture_open_io_method(us_capture_s *cap);
+ static int _capture_open_io_method_mmap(us_capture_s *cap);
+ static int _capture_open_io_method_userptr(us_capture_s *cap);
+@@ -144,7 +146,7 @@ us_capture_s *us_capture_init(void) {
+ 	cap->format = V4L2_PIX_FMT_YUYV;
+ 	cap->jpeg_quality = 80;
+ 	cap->standard = V4L2_STD_UNKNOWN;
+-	cap->io_method = V4L2_MEMORY_MMAP;
++	cap->io_method = V4L2_MEMORY_USERPTR;
+ 	cap->n_bufs = us_get_cores_available() + 1;
+ 	cap->min_frame_size = 128;
+ 	cap->timeout = 1;
+@@ -185,6 +187,27 @@ int us_capture_parse_io_method(const char *str) {
+ }
+ 
+ int us_capture_open(us_capture_s *cap) {
++	const enum v4l2_memory requested_io_method = cap->io_method;
++	int result = _capture_open_impl(cap);
++
++	if (result >= 0) {
++		return result;
++	}
++
++	if (requested_io_method != V4L2_MEMORY_USERPTR) {
++		return result;
++	}
++
++	_LOG_INFO("USERPTR capture failed, retrying with MMAP");
++	cap->io_method = V4L2_MEMORY_MMAP;
++	result = _capture_open_impl(cap);
++	if (result < 0) {
++		cap->io_method = requested_io_method;
++	}
++	return result;
++}
++
++static int _capture_open_impl(us_capture_s *cap) {
+ 	us_capture_runtime_s *const run = cap->run;
+ 
+ 	if (access(cap->path, R_OK | W_OK) < 0) {
+@@ -1009,12 +1032,11 @@ static int _capture_open_io_method_userptr(us_capture_s *cap) {
+ 	US_CALLOC(run->bufs, req.count);
+ 
+ 	const uint page_size = getpagesize();
+-	const uint buf_size = us_align_size(run->raw_size, page_size);
++	const uint buf_size = _capture_get_userptr_buffer_size(run);
+ 
+ 	for (run->n_bufs = 0; run->n_bufs < req.count; ++run->n_bufs) {
+ 		us_capture_hwbuf_s *hw = &run->bufs[run->n_bufs];
+ 		US_A((hw->raw.data = aligned_alloc(page_size, buf_size)) != NULL);
+-		memset(hw->raw.data, 0, buf_size);
+ 		hw->raw.allocated = buf_size;
+ 		if (run->capture_mplane) {
+ 			US_CALLOC(hw->buf.m.planes, VIDEO_MAX_PLANES);
+@@ -1023,6 +1045,15 @@ static int _capture_open_io_method_userptr(us_capture_s *cap) {
+ 	return 0;
+ }
+ 
++static uz _capture_get_userptr_buffer_size(const us_capture_runtime_s *run) {
++	// The driver-reported sizeimage is much larger than the actual MJPEG payload
++	// we observe on this target. Cap USERPTR to keep the resident capture buffer
++	// small while still leaving a conservative safety margin.
++	const uz raw_size = run->raw_size;
++	const uz cap_size = 256 * 1024UL;
++	return us_align_size(US_MIN(raw_size, cap_size), getpagesize());
++}
++
+ static int _capture_open_queue_buffers(us_capture_s *cap) {
+ 	us_capture_runtime_s *const run = cap->run;
+ 
+diff --git a/src/libs/frame.c b/src/libs/frame.c
+index a770969e9ad7134e2c98328c9b906782090aafcf..222eb06cd7e4e803d429c6c54c4152d2f1a75acf 100644
+--- a/src/libs/frame.c
++++ b/src/libs/frame.c
+@@ -35,27 +35,67 @@
+ us_frame_s *us_frame_init(void) {
+ 	us_frame_s *frame;
+ 	US_CALLOC(frame, 1);
+-	us_frame_realloc_data(frame, 32 * 1024);
++	// Most live JPEG buffers settle well below 32 KiB in this deployment,
++	// so keep the startup floor small and grow only when needed.
++	us_frame_realloc_data(frame, 8 * 1024);
+ 	frame->dma_fd = -1;
+ 	return frame;
+ }
+ 
+ void us_frame_destroy(us_frame_s *frame) {
+-	US_DELETE(frame->data, free);
++	if (!frame->external_data) {
++		US_DELETE(frame->data, free);
++	}
+ 	free(frame);
+ }
+ 
+ void us_frame_realloc_data(us_frame_s *frame, uz size) {
++	if (frame->external_data) {
++		u8 *owned;
++		US_CALLOC(owned, size);
++		if (frame->data != NULL && frame->used > 0) {
++			memcpy(owned, frame->data, frame->used);
++		}
++		frame->data = owned;
++		frame->external_data = false;
++		frame->allocated = size;
++		return;
++	}
+ 	if (frame->allocated < size) {
+ 		US_REALLOC(frame->data, size);
+ 		frame->allocated = size;
+ 	}
+ }
+ 
++void us_frame_release_data(us_frame_s *frame) {
++	if (frame->external_data) {
++		frame->data = NULL;
++		frame->used = 0;
++		frame->allocated = 0;
++		frame->external_data = false;
++		return;
++	}
++
++	US_DELETE(frame->data, free);
++	frame->used = 0;
++	frame->allocated = 0;
++}
++
++void us_frame_set_external_data(us_frame_s *frame, const u8 *data, uz size) {
++	if (!frame->external_data) {
++		US_DELETE(frame->data, free);
++	}
++	frame->data = (u8*)data;
++	frame->used = size;
++	frame->allocated = 0;
++	frame->external_data = true;
++}
++
+ void us_frame_set_data(us_frame_s *frame, const u8 *data, uz size) {
+ 	us_frame_realloc_data(frame, size);
+ 	memcpy(frame->data, data, size);
+ 	frame->used = size;
++	frame->external_data = false;
+ }
+ 
+ void us_frame_append_data(us_frame_s *frame, const u8 *data, uz size) {
+diff --git a/src/libs/frame.h b/src/libs/frame.h
+index 74c45512c6fa1fd0755b621483294ffb61e0292d..a24ce4bf994edf24748ed1e1cbc0e736f0641df7 100644
+--- a/src/libs/frame.h
++++ b/src/libs/frame.h
+@@ -49,6 +49,7 @@ typedef struct {
+ 	uz		used;
+ 	uz		allocated;
+ 	int		dma_fd;
++	bool	external_data;
+ 
+ 	US_FRAME_META_DECLARE;
+ } us_frame_s;
+@@ -102,6 +103,8 @@ us_frame_s *us_frame_init(void);
+ void us_frame_destroy(us_frame_s *frame);
+ 
+ void us_frame_realloc_data(us_frame_s *frame, uz size);
++void us_frame_release_data(us_frame_s *frame);
++void us_frame_set_external_data(us_frame_s *frame, const u8 *data, uz size);
+ void us_frame_set_data(us_frame_s *frame, const u8 *data, uz size);
+ void us_frame_append_data(us_frame_s *frame, const u8 *data, uz size);
+ 
+diff --git a/src/libs/frametext.c b/src/libs/frametext.c
+index d5adfc73765558dd3f7e0b5052dec9c4678b1041..9daa61f02666a588c3d63378aeede9d1faee6393 100644
+--- a/src/libs/frametext.c
++++ b/src/libs/frametext.c
+@@ -92,6 +92,7 @@ void us_frametext_draw(us_frametext_s *ft, const char *text, uint width, uint he
+ 	if (
+ 		frame->width == width && frame->height == height
+ 		&& ft->text != NULL && !strcmp(ft->text, text)
++		&& frame->data != NULL && frame->used > 0
+ 	) {
+ 		return;
+ 	}
+diff --git a/src/libs/queue.c b/src/libs/queue.c
+index d72f656a7fdfde16c6e7f83202f8914a8cd808c7..854ecb11556a441b632a8365886e16ca9c22f5e1 100644
+--- a/src/libs/queue.c
++++ b/src/libs/queue.c
+@@ -107,5 +107,5 @@ bool us_queue_is_empty(us_queue_s *q) {
+ 	US_MUTEX_LOCK(q->mutex);
+ 	const uint size = q->size;
+ 	US_MUTEX_UNLOCK(q->mutex);
+-	return (bool)(q->capacity - size);
++	return (size == 0);
+ }
+diff --git a/src/libs/threading.h b/src/libs/threading.h
+index 2f161af24c9c78702a6eaca2ad6438f3df7783f2..be145e1407259bf9a9088ecb8295034248cd0988 100644
+--- a/src/libs/threading.h
++++ b/src/libs/threading.h
+@@ -46,7 +46,41 @@
+ #	define US_THREAD_NAME_SIZE ((uz)16)
+ #endif
+ 
+-#define US_THREAD_CREATE(x_tid, x_func, x_arg)	US_A(!pthread_create(&(x_tid), NULL, (x_func), (x_arg)))
++#ifndef US_THREAD_STACK_SIZE
++#	define US_THREAD_STACK_SIZE (32 * 1024UL)
++#endif
++
++INLINE int us_thread_create(pthread_t *tid, void *(*func)(void*), void *arg) {
++#if (US_THREAD_STACK_SIZE > 0)
++	pthread_attr_t attr;
++	US_A(!pthread_attr_init(&attr));
++
++	// These threads are long-lived but have shallow call stacks, so cap the
++	// default pthread stack reservation to keep memory overhead predictable.
++	uz stack_size = US_THREAD_STACK_SIZE;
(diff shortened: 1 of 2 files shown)


pdscomp:

- `0002-reduce-memory-usage.patch`: very large patch, will take significant time to understand.
- `ustreamer-init-d`: disagrees with 10→8 FPS drop; in his testing earlier FPS didn't see a difference.
- Patch looks untargeted; goal per-patch unclear.

pdscomp:
Thanks for the review. I agree this patch is broad and exploratory in places, so I want to clarify the goal and the risk profile.

uStreamer is a non-essential printer component, and rollback is trivial: the service can be disabled or the recipe patch can be reverted without affecting printing, Klipper, or Moonraker. That makes it a reasonable place to try several narrowly targeted memory reductions while we have a reproducible CC1 workload.

The common goal for every change here is reducing uStreamer’s PSS and RSS. During the investigation, the individual changes generally moved memory usage by roughly 100 KiB or more, and the combined result was:

- PSS: 4.67 MiB → 3.08 MiB
- RSS: 5.76 MiB → 4.14 MiB

The main finding was that uStreamer’s capture/processing pipeline retained multiple copies of frame buffers for throughput. The camera is already configured for low resolution and roughly 10 FPS, so that optimization is a poor trade on a 128 MiB appliance. The kernel/V4L2 capture path also reserves approximately 1.8 MiB per buffer even though the camera’s compressed MJPEG frames are usually only 20–30 KiB. Retaining several of those buffers therefore costs more than 3 MiB of scarce RAM for very little practical benefit. Keeping only the newest frame and allowing HTTP delivery to borrow it safely is where much of the improvement comes from.

The patch also keeps the camera-specific fallback behavior: the CC1 UVC device rejects USERPTR queueing, so the implementation retries with MMAP and remains functional. Snapshots and live video were verified on the target.

Based on this feedback, I reverted the 10→8 FPS change. The memory reduction patch can therefore be evaluated independently while retaining the existing 10 FPS deployment setting.

If useful, I can further split the C patch into smaller commits, but the changes are all aimed at the same measurable objective: reducing uStreamer’s resident memory footprint on the CC1.

pdscomp:
Flashed and tested `PR-255-1d7` on `carbon2u` (CC1).

ustreamer is running and serving snapshots + live stream:
```text
/usr/bin/ustreamer -r 864x480 -c HW -m MJPEG -b 1 -w 1 -f 10 -l --host=0.0.0.0 --allow-origin=*
```

Measured from `/proc/[pid]/smaps_rollup`:

| state | PSS | RSS |
|---|---|---:|
| idle | 2.88 MiB | 4.02 MiB |
| after `/snapshot` | 3.13 MiB | 4.31 MiB |
| after 2s `/stream` | 2.62 MiB | 3.80 MiB |

- `/snapshot` returns HTTP 200 with ~51 kB JPEG.
- `/stream` delivers MJPEG successfully (pulled ~900 kB in 2s).
- No crashes or elevated memory after the stream.

These are absolute numbers from the patched build `PR-255-1d7`, not a before/after baseline. They line up with the PR body's claim of ~3.08 MiB PSS / 4.14 MiB RSS at 8 FPS; observed at 10 FPS is slightly lower.

TheDevMinerTV (commented):
> [!NOTE]
> This review was produced by an LLM. Findings below were reproduced empirically against builds of the pinned `SRCREV`, not just read off the diff — the repro commands are included. Please still sanity-check anything that looks off.

## Summary

**Recommend: request changes.** Three blockers, all reproduced. The core idea — reclaiming the blank RGB canvas — is sound and is where ~75% of the claimed saving actually comes from, but it is implemented unconditionally and crashes a supported path. Two of the five bundled changes (USERPTR, thread stacks) are provably worth nothing on this target.

### Method

Cloned `pikvm/ustreamer` at the pinned `SRCREV=88460b72e191035d04355e25106af817cbfe069e`, built two binaries — baseline (`0001` only) and patched (`0001` + `0002`) — and A/B'd them with the production argv from `ustreamer-init-d` (`-r 864x480 -c HW -m MJPEG -b 1 -w 1 -f 10`). Both patches apply cleanly in order; patched builds warning-free. Host is x86-64/glibc, so absolute numbers differ from the CC1, but the deltas and the crashes are architecture-independent. The USERPTR behaviour was probed against a real UVC camera with a standalone `S_FMT`+`REQBUFS`+`QBUF` program (no `STREAMON`, no frames captured).

Not verified: the live streaming path against a camera, and on-target ARM numbers.

## Blockers

**1. `--h264-sink` → `SIGABRT`.** `us_frame_release_data(blank->raw)` frees a buffer that has three live consumers. One asserts:

```
ustreamer: ustreamer/../libs/frame.h:88: us_frame_encoding_begin: Assertion `m_ar && "src->used > 0"' failed.
patched: DIED rc=134   (core dumped)
base:    ALIVE
```

Repro: `ustreamer -d /dev/nonexistent0 -r 864x480 -c HW -m MJPEG -b 1 -w 1 -f 10 --h264-sink=us.pat.h264 --h264-sink-rm`. Fires on any camera error or absence. `US_A` is a plain `assert()` and the Yocto recipe does not set `-DNDEBUG`, so it is live in the shipped build.

**2. `--raw-sink` publishes empty frames.** Same root cause:

```
base raw dump:    rawbase.bin  8709120 bytes
patched raw dump: rawpat.bin         0 bytes
```

**3. The USERPTR change is dead code, and the commit message misdiagnoses why.** Details inline at the `256 * 1024UL` line. Short version: UVC accepts USERPTR; the patch's own cap is what makes `QBUF` return `EINVAL`. Net memory effect on this target is exactly zero, and it adds unbounded log spam plus a doubled open cost whenever the camera is absent.

## Major

**4. Offline steady state is measurably worse.** Device absent, production argv, three samples 12s apart, all stable:

| | base | patched | delta |
|---|---:|---:|---:|
| PSS | 2768 kB | **3530 kB** | +762 kB |
| RSS | 5812 kB | **6640 kB** | +828 kB |
| VSZ | 89564 kB | 72844 kB | −16720 kB |

`us_frame_release_data` `free()`s the 1.24 MiB canvas, then `us_frametext_draw` re-`malloc`s + `memset`s + re-renders it every second. On glibc, freeing an mmap'd 1.24 MiB block raises `mmap_threshold`, so subsequent allocations come from the main arena and are never returned. Offline you now pay more memory *and* a full text re-render + JPEG compress per second — which is exactly the state a printer is in when its USB camera is flaky.

**5. The VSZ delta above is the entire thread-stack effect** — 16720 kB is precisely 2 × 8 MiB of default stack reservation, all virtual. RSS went up, not down. See inline.

**6. `jpeg_ring` 4→1 plus the `hwbuf_decref` reorder** — see inline. Not reproduced under load.

## Suggestion

Split this. Land the blank-canvas reclaim alone, gated (`raw_sink == NULL && h264_sink == NULL`) or made lazy so consumers re-render on demand — that is the real 1.24 MiB. Drop the USERPTR default + cap and the stack cap entirely. Send the `us_queue_is_empty` fix upstream on its own; it is a genuine bug fix that does not belong in a memory patch. `jpeg_ring` 4→2 is defensible, 1 needs load testing.

## Docs

PR body says CC1 / after 3.08 PSS / 4.14 RSS; the commit message says CC2 / 3.17 / 4.30. The title says 8 FPS; `ustreamer-init-d` ships `-f 10`. Worth reconciling before this lands, since the numbers are the justification.

meta-opencentauri/recipes-apps/ustreamer/files/0002-reduce-memory-usage.patch
```diff
@@ -0,0 +1,371 @@
...
+
+* cap internal thread stacks;
+* retain only the newest JPEG in the streamer backlog and let HTTP clients
+  borrow a frame while it remains valid;
+* avoid allocating a permanent RGB canvas for the blank frame;
+* start generic frame buffers smaller and grow them on demand; and
+* use bounded USERPTR buffers when supported, with an MMAP retry when a UVC
+  camera rejects USERPTR.
+
+The CC2 UVC camera rejects USERPTR queueing, so MMAP remains the validated
```
TheDevMinerTV, line 22: **This is a misdiagnosis, and it matters.** The UVC camera does not reject USERPTR — the 256 KiB cap added by this same patch is what makes `QBUF` fail.

Probed against a real UVC camera (`S_FMT` + `REQBUFS` + `QBUF` only, no `STREAMON`, no frames captured):

```
negotiated 848x480 sizeimage=814080
REQBUFS(USERPTR) OK, count=1
QBUF len=262144 FAILED: Invalid argument
QBUF len=817152 OK
```

`REQBUFS(USERPTR)` succeeds. `QBUF` at full `sizeimage` succeeds. `QBUF` at the capped 256 KiB fails, because vb2's `__verify_length()` rejects `length < sizeimage`. So the fallback that this paragraph presents as evidence of camera limitation is really the patch tripping over its own cap.

meta-opencentauri/recipes-apps/ustreamer/files/0002-reduce-memory-usage.patch
```diff
@@ -0,0 +1,371 @@
...
++static int _capture_open_impl(us_capture_s *cap);
+ static int _capture_open_io_method(us_capture_s *cap);
+ static int _capture_open_io_method_mmap(us_capture_s *cap);
+ static int _capture_open_io_method_userptr(us_capture_s *cap);
+@@ -144,7 +146,7 @@ us_capture_s *us_capture_init(void) {
+ 	cap->format = V4L2_PIX_FMT_YUYV;
+ 	cap->jpeg_quality = 80;
+ 	cap->standard = V4L2_STD_UNKNOWN;
+-	cap->io_method = V4L2_MEMORY_MMAP;
++	cap->io_method = V4L2_MEMORY_USERPTR;
```
TheDevMinerTV, line 60: Flipping the upstream default for every user, to reach a path that cannot succeed at this resolution (see the cap at line 115). Since `sizeimage` is 814080 at 864x480 and the cap is 262144, `QBUF` returns `EINVAL` every time, so this always falls back to MMAP.

Cost of the flip: two full open attempts instead of one on every start and every reconnect, plus the log spam at line 79. Benefit: zero. Recommend dropping the USERPTR work from this PR entirely.

meta-opencentauri/recipes-apps/ustreamer/files/0002-reduce-memory-usage.patch
```diff
@@ -0,0 +1,371 @@
...
++
++	if (result >= 0) {
++		return result;
++	}
++
++	if (requested_io_method != V4L2_MEMORY_USERPTR) {
++		return result;
++	}
++
++	_LOG_INFO("USERPTR capture failed, retrying with MMAP");
```
TheDevMinerTV, line 79: **Unbounded INFO spam when the camera is absent.** This bypasses the `US_ONCE_FOR(run->open_error_once, ...)` de-duplication upstream uses, and the retry fires on every error class including `US_ERROR_NO_DEVICE` — where retrying with a different IO method cannot possibly help.

Measured over ~10s with the device missing:

```
base log lines:    9
patched log lines: 27   (18x "CAP: USERPTR capture failed, retrying with MMAP", 1/sec, forever)
```

On a 128 MiB box this floods syslog for as long as the camera stays unplugged. If the retry survives review, at minimum gate it on `result == US_ERROR_COMMON` and log at DEBUG, or once.

meta-opencentauri/recipes-apps/ustreamer/files/0002-reduce-memory-usage.patch
```diff
@@ -0,0 +1,371 @@
...
+@@ -1023,6 +1045,15 @@ static int _capture_open_io_method_userptr(us_capture_s *cap) {
+ 	return 0;
+ }
+ 
++static uz _capture_get_userptr_buffer_size(const us_capture_runtime_s *run) {
++	// The driver-reported sizeimage is much larger than the actual MJPEG payload
++	// we observe on this target. Cap USERPTR to keep the resident capture buffer
++	// small while still leaving a conservative safety margin.
++	const uz raw_size = run->raw_size;
++	const uz cap_size = 256 * 1024UL;
```
TheDevMinerTV, line 115: **This cap makes USERPTR unusable, which is why it never engages.** `_capture_open_queue_buffers` passes `buf.length = run->bufs[i].raw.allocated`, and vb2 rejects `length < sizeimage` for USERPTR. At 864x480 MJPEG the driver reports `sizeimage=814080`, so 262144 always returns `EINVAL` — verified on real UVC hardware (see my note on line 22).

Two consequences:

1. On this target the entire USERPTR change contributes **0 bytes** of savings. It is pure startup cost and log noise.
2. On a hypothetical camera that negotiates `sizeimage <= 256 KiB`, USERPTR *would* engage — into a path that has never been exercised on this device, and where this patch has also removed the `memset(hw->raw.data, 0, buf_size)` zeroing just above. `bytesused` should bound what is read, but this is untested code shipping to production.

The comment claims the driver-reported `sizeimage` is "much larger than the actual MJPEG payload". True, but irrelevant — the driver validates against `sizeimage`, not against payload size. There is no way to shrink a USERPTR buffer below `sizeimage`.

meta-opencentauri/recipes-apps/ustreamer/files/0002-reduce-memory-usage.patch
```diff
@@ -0,0 +1,371 @@
...
+--- a/src/libs/frame.c
++++ b/src/libs/frame.c
+@@ -35,27 +35,67 @@
+ us_frame_s *us_frame_init(void) {
+ 	us_frame_s *frame;
+ 	US_CALLOC(frame, 1);
+-	us_frame_realloc_data(frame, 32 * 1024);
++	// Most live JPEG buffers settle well below 32 KiB in this deployment,
++	// so keep the startup floor small and grow only when needed.
++	us_frame_realloc_data(frame, 8 * 1024);
```
TheDevMinerTV, line 133: This one is fine on its own merits — worth roughly 24 KiB x ~10 frames, and the frames grow on demand anyway. No objection; just noting it is one of the few changes here that does what it says without side effects.

meta-opencentauri/recipes-apps/ustreamer/files/0002-reduce-memory-usage.patch
```diff
@@ -0,0 +1,371 @@
...
++	}
+ 	free(frame);
+ }
+ 
+ void us_frame_realloc_data(us_frame_s *frame, uz size) {
++	if (frame->external_data) {
++		u8 *owned;
++		US_CALLOC(owned, size);
++		if (frame->data != NULL && frame->used > 0) {
++			memcpy(owned, frame->data, frame->used);
```
TheDevMinerTV, line 151: **Missing length clamp — latent heap overflow.** `owned` is `size` bytes, but this copies `frame->used` bytes. If any caller ever reaches `us_frame_realloc_data(frame, size)` on an external frame with `used > size`, this overflows.

Today it is safe *only* because `us_frame_encoding_begin` sets `dest->used = 0` before `_copy_plus_huffman` calls `us_frame_set_data`. That is an invariant living in a different file, undocumented, and one refactor away from a memory-corruption bug.

```c
memcpy(owned, frame->data, US_MIN(frame->used, size));
```

Worth doing regardless of what happens to the rest of the patch.

meta-opencentauri/recipes-apps/ustreamer/files/0002-reduce-memory-usage.patch
```diff
@@ -0,0 +1,371 @@
...
+diff --git a/src/libs/queue.c b/src/libs/queue.c
+index d72f656a7fdfde16c6e7f83202f8914a8cd808c7..854ecb11556a441b632a8365886e16ca9c22f5e1 100644
+--- a/src/libs/queue.c
++++ b/src/libs/queue.c
+@@ -107,5 +107,5 @@ bool us_queue_is_empty(us_queue_s *q) {
+ 	US_MUTEX_LOCK(q->mutex);
+ 	const uint size = q->size;
+ 	US_MUTEX_UNLOCK(q->mutex);
+-	return (bool)(q->capacity - size);
++	return (size == 0);
```
TheDevMinerTV, line 238: This is a **correct bug fix** but it does not belong in a memory-reduction patch. Upstream's `return (bool)(q->capacity - size)` evaluates to "is not full", which also silently breaks `US_QUEUE_DELETE_WITH_ITEMS` in `queue.h` — its drain loop only runs while the queue is full, so items leak on destroy.

Please send it to pikvm/ustreamer as its own PR and carry it as a separate `.patch` file. Bundling it here means the crash in `blank.c` cannot be reverted without also reverting this.

One side effect to be aware of: this makes the drain loop in `_get_latest_hw` (`stream.c:528`) run whenever `size > 1` rather than almost never, which exposes a TOCTOU between `us_queue_is_empty` and `US_A(!us_queue_get(q, ..., 0))` — the lock is dropped in between. Safe today because that queue is single-consumer, but the assert would abort if that ever changes.

meta-opencentauri/recipes-apps/ustreamer/files/0002-reduce-memory-usage.patch
```diff
@@ -0,0 +1,371 @@
...
+index 2f161af24c9c78702a6eaca2ad6438f3df7783f2..be145e1407259bf9a9088ecb8295034248cd0988 100644
+--- a/src/libs/threading.h
++++ b/src/libs/threading.h
+@@ -46,7 +46,41 @@
+ #	define US_THREAD_NAME_SIZE ((uz)16)
+ #endif
+ 
+-#define US_THREAD_CREATE(x_tid, x_func, x_arg)	US_A(!pthread_create(&(x_tid), NULL, (x_func), (x_arg)))
++#ifndef US_THREAD_STACK_SIZE
++#	define US_THREAD_STACK_SIZE (32 * 1024UL)
```
TheDevMinerTV, line 250: **This buys VSZ, not RSS.** Untouched stack pages are never resident, so capping the reservation does not reduce PSS. Measured, device absent, production argv:

| | base | patched |
|---|---:|---:|
| VSZ | 89564 kB | 72844 kB |
| RSS | 5812 kB | **6640 kB** |
| PSS | 2768 kB | **3530 kB** |

The 16720 kB VSZ drop is exactly 2 x 8 MiB of default stack reservation. RSS and PSS both went *up*. Linux overcommits by default, so the VSZ reduction is not worth anything on the CC1 either.

Against that zero benefit: the JPEG worker thread runs libjpeg-turbo on 32 KiB minus static TLS (glibc places the static TLS block inside the thread stack allocation) minus the guard page. There is a guard page so an overflow is a clean `SIGSEGV` rather than corruption, but a crash under an unusual input is still a crash. Also note this applies to *every* `US_THREAD_CREATE` in the codebase, not just the ones profiled here.

Recommend dropping this hunk.

meta-opencentauri/recipes-apps/ustreamer/files/0002-reduce-memory-usage.patch
```diff
@@ -0,0 +1,371 @@
...
+ #ifdef WITH_PTHREAD_NP
+diff --git a/src/ustreamer/blank.c b/src/ustreamer/blank.c
+index 09b1b0008c406d43ffba1d78b7f1664b749d63ea..50fb5af947acabeee2b78e8be9746d3ab8a13b43 100644
+--- a/src/ustreamer/blank.c
++++ b/src/ustreamer/blank.c
+@@ -43,6 +43,7 @@ us_blank_s *us_blank_init(void) {
+ void us_blank_draw(us_blank_s *blank, const char *text, uint width, uint height) {
+ 	us_frametext_draw(blank->ft, text, width, height);
+ 	us_cpu_encoder_compress(blank->raw, blank->jpeg, 95);
++	us_frame_release_data(blank->raw);
```
TheDevMinerTV, line 295: **Blocker.** `blank->raw` is `blank->ft->frame`, and it has three consumers after this call, in `stream.c:663-666`:

```c
_stream_update_captured_fpsi(stream, run->blank->raw, false);   // meta only, OK
_stream_expose_jpeg(stream, run->blank->jpeg);                  // OK, jpeg already compressed
_stream_expose_raw(stream, run->blank->raw);                    // publishes used=0
_stream_encode_expose_h264(stream, run->blank->raw, true);      // ABORTS
```

**`--h264-sink` crashes.** `_stream_encode_expose_h264` -> `us_m2m_encoder_compress` -> `us_frame_encoding_begin`, which asserts `src->used > 0` before it does anything else:

```
ustreamer: ustreamer/../libs/frame.h:88: us_frame_encoding_begin: Assertion `m_ar && "src->used > 0"' failed.
patched: DIED rc=134   (core dumped)
base:    ALIVE
```

Repro: `-d /dev/nonexistent0 --h264-sink=us.pat.h264 --h264-sink-rm`, wait ~5s.

**`--raw-sink` silently publishes nothing.** `us_memsink_server_put` copies `frame->used == 0` bytes:

```
base raw dump:    8709120 bytes
patched raw dump:       0 bytes
```

Neither sink is in the current `ustreamer-init-d`, so this is not production-breaking today — but the PR describes the risk as "Low", and this is a reproducible abort in a path upstream supports.

Worth keeping the idea though: this canvas is 864*480*3 = 1.19 MiB, which is roughly 75% of the 1.59 MiB the PR reports. It just needs to be conditional (`raw_sink == NULL && h264_sink == NULL`) or lazy, so consumers can force a re-render. See also my note on line 4 of the summary about the offline-state regression this introduces on glibc.

meta-opencentauri/recipes-apps/ustreamer/files/0002-reduce-memory-usage.patch
```diff
@@ -0,0 +1,371 @@
...
+--- a/src/ustreamer/encoders/hw/encoder.c
++++ b/src/ustreamer/encoders/hw/encoder.c
+@@ -69,7 +69,9 @@ void _copy_plus_huffman(const us_frame_s *src, us_frame_s *dest) {
+ 		us_frame_append_data(dest, src_ptr, src->used - paste);
+ 
+ 	} else {
+-		us_frame_set_data(dest, src->data, src->used);
++		// Borrow the camera JPEG buffer directly when it is already complete.
++		// The caller keeps the capture buffer alive until after exposure.
++		us_frame_set_external_data(dest, src->data, src->used);
```
TheDevMinerTV, line 310: The borrow itself is **correct**, given the `decref` reorder at line 367 — `_stream_expose_jpeg` does a real `us_frame_copy` into the ring slot before the hwbuf is requeued.

Two things worth pinning down before this lands:

1. The comment says "The caller keeps the capture buffer alive until after exposure." That is true only because of the reorder in a different file. Please state the coupling explicitly in both places, or the next person to touch `_jpeg_thread` will reintroduce a use-after-requeue.
2. On the `job_failed` / `!job_timely` drop paths, `dest` keeps a dangling pointer with `used > 0` and `external_data = true` after the `decref`. Nothing dereferences it today (the next `us_frame_encoding_begin` zeroes `used` first), but combined with the missing clamp at line 151 that is two accidents away from reading requeued driver memory. Consider calling `us_frame_release_data(job->dest)` on the drop paths to make the invariant explicit rather than incidental.

meta-opencentauri/recipes-apps/ustreamer/files/0002-reduce-memory-usage.patch
```diff
@@ -0,0 +1,371 @@
...
+diff --git a/src/ustreamer/options.c b/src/ustreamer/options.c
+index 048e6c50854edbf369c39e9ca4128205364d4993..2d36e1277c0b6ba0768562f41f2da7c52f0e7ae4 100644
+--- a/src/ustreamer/options.c
++++ b/src/ustreamer/options.c
+@@ -703,7 +703,7 @@ static void _help(
+ 	SAY("                                           Available: %s; default: disabled.\n", US_STANDARDS_STR);
+ 	SAY("    -I|--io-method <method>  ───────────── Set V4L2 IO method (see kernel documentation).");
+ 	SAY("                                           Changing of this parameter may increase the performance. Or not.");
+-	SAY("                                           Available: %s; default: MMAP.\n", US_IO_METHODS_STR);
++	SAY("                                           Available: %s; default: USERPTR, with MMAP fallback if needed.\n", US_IO_METHODS_STR);
```
TheDevMinerTV, line 323: Documents a changed upstream default that this deployment never exercises — `-I` is not passed in `ustreamer-init-d`, and the USERPTR path cannot succeed anyway (line 115). If the USERPTR hunks are dropped as suggested, this goes with them.

meta-opencentauri/recipes-apps/ustreamer/files/0002-reduce-memory-usage.patch
```diff
@@ -0,0 +1,371 @@
...
+diff --git a/src/ustreamer/stream.c b/src/ustreamer/stream.c
+index 9ae94e42b35d355df43d449a595810eefb06d501..dd28e3e64589d916e96f37d67fbeb0feaf8f3be8 100644
+--- a/src/ustreamer/stream.c
++++ b/src/ustreamer/stream.c
+@@ -58,6 +58,11 @@
+ #endif
+ 
+ 
++#ifndef US_HTTP_JPEG_RING_SIZE
++#	define US_HTTP_JPEG_RING_SIZE 1
```
TheDevMinerTV, line 336: A capacity of 1 removes all pipelining between the encoder and the HTTP exposer, turning a buffered handoff into a rendezvous.

The wait on the producer side is a **busy spin with no sleep**:

```c
while ((ri = us_ring_producer_acquire(run->http->jpeg_ring, 0)) < 0) {
    if (atomic_load(&run->stop)) { return; }
}
```

`timeout = 0` means `us_queue_get` -> `pthread_cond_timedwait` returns `ETIMEDOUT` immediately, so this pegs a core for as long as `_http_refresher` holds the only slot. Upstream's 4 slots make that window essentially unreachable; at 1 it is reachable any time the libevent loop is busy writing to a slow client.

Combined with the `decref` reorder at line 367 and production's `-b 1`, the spin now happens while holding the *only* capture buffer, so capture stalls for its duration. At 10 FPS there is 100ms of slack so this may well never bite — but I could not exercise it without a camera, and it is the kind of thing that only shows up under a stalled HTTP client in the field.

Savings here are roughly 3 x ~50 KiB. 4->2 keeps a real buffer and most of the win.

meta-opencentauri/recipes-apps/ustreamer/files/0002-reduce-memory-usage.patch
```diff
@@ -0,0 +1,371 @@
...
+-			us_capture_hwbuf_decref(job->hw);
+-			job->hw = NULL;
+ 			if (wr->job_failed) {
+ 				// pass
+ 			} else if (wr->job_timely) {
+@@ -345,6 +350,8 @@ static void *_jpeg_thread(void *v_ctx) {
+ 			} else {
+ 				US_LOG_PERF("JPEG: ----- Encoded JPEG dropped; worker=%s", wr->name);
+ 			}
++			us_capture_hwbuf_decref(job->hw);
```
TheDevMinerTV, line 367: This reorder is **required** for the `us_frame_set_external_data` borrow at line 310 to be safe, and it is done correctly — the `decref` now happens after `_stream_expose_jpeg` has copied the frame into the ring slot.

The cost is that the capture buffer is held across `_stream_expose_jpeg`, which includes the unbounded busy-spin on `us_ring_producer_acquire` (see line 336) plus a full JPEG memcpy. With production's `-b 1` that is the only buffer, so anything that delays the exposer now directly stalls capture. Fine at 4 ring slots, riskier at 1 — another reason not to take both changes together.

PR #257, opened 2026-07-19
python: lazy-load optional NumPy modules

pdscomp:
# python: reduce Klipper/Moonraker application memory overhead

## What changed

This branch keeps the previously validated `-X no_debug_ranges` service flags
and adds a targeted NumPy/Kalico memory reduction for the 128 MiB CC1:

- Backports NumPy 2.0's lazy public-submodule loading pattern to pinned NumPy
  1.26.4 for `fft`, `random`, `polynomial`, `ma`, `ctypeslib`, and `linalg`.
- Makes Kalico temperature-fan NumPy usage lazy; PID and watermark fans no
  longer import NumPy at module startup.
- Removes the eager NumPy dependency from Kalico's JSON conversion fallback;
  NumPy values are recognized by type metadata only when conversion is needed.

The full test log is available as a [public artifact](https://gist.github.com/pdscomp/a062755da204c914c072de3391ea6c1d).

## Validation

- Full CC1 image build passed all 6,521 BitBake tasks.
- SWU flashed successfully to `carbon2u` and rebooted cleanly.
- Wi-Fi re-associated; Klipper, Moonraker, camera, I²C, USB, and all tested
  APIs remained healthy.
- Target import testing confirmed five of the six targeted NumPy subpackages
  stay unloaded after `import numpy`; explicit `numpy.linalg` and
  `numpy.random` access still works. `numpy.linalg` is imported indirectly by
  NumPy 1.26 matrixlib and is documented as a remaining limitation.
- Real X-axis `SHAPER_CALIBRATE` completed through NumPy analysis and returned
  `mzv` at 55.4 Hz with 0% measured vibrations.
- The production full-calibration macro completed X/Y input shaping, extruder
  PID (`Kp=27.781`, `Ki=2.590`, `Kd=74.488`), and load-cell tare calibration.
- The bed mesh reached the known pre-existing load-cell drift failure; this is
  not attributed to the Python/NumPy changes. Heater targets were cleared and
  no configuration was saved automatically.

## Experiments not retained

These were tested and either regressed memory, changed semantics, or did not
produce enough repeatable benefit to justify the complexity:

- Moonraker `-O`: removed runtime assertions and used about 3.2 MiB more
  effective memory after a clean reboot.
- Matching precompiled `.pyc`: no steady-state or startup benefit over
  `no_debug_ranges`, while increasing the image and update complexity.
- `MALLOC_ARENA_MAX=2`: no repeatable settled-idle gain.
- `python -S`: only about 0.27 MiB for a bare interpreter and bypasses `.pth`
  processing.
- Disabling uvloop: the packaged extension is not active on the target, so it
  offers no runtime-memory benefit.
- Removing application components or either static UI: storage-only changes,
  not meaningful resident-RAM savings.

## Scope and rollback

No printer configuration is saved by the test procedure. The patch is
reversible by removing the two recipe patches and the NumPy bbappend. The
known mesh drift issue remains a separate load-cell investigation.

### DIFF
--- a/meta-opencentauri/recipes-apps/klipper/files/0001-Lazy-load-optional-NumPy-users.patch
+++ b/meta-opencentauri/recipes-apps/klipper/files/0001-Lazy-load-optional-NumPy-users.patch
@@ -1,0 +1,107 @@
+From 7f1b2a37d5cb51aabbec6483588f6a58c91bcf31 Mon Sep 17 00:00:00 2001
+From: OpenCentauri <dev@opencentauri.io>
+Date: Mon, 20 Jul 2026 15:00:00 -0400
+Subject: [PATCH] Lazy-load optional NumPy users
+
+Upstream-Status: Pending
+
+Importing NumPy 1.26 brings a substantial set of numerical modules into the
+Klippy process.  Do not make every configuration pay that cost merely for JSON
+conversion or for temperature-fan control modes which do not use curves.
+
+Recognize NumPy values by their type metadata in the JSON fallback, which is
+only invoked for otherwise non-serializable objects.  Import NumPy from the
+curve controller itself so watermark and PID temperature fans remain free of
+that dependency.
+---
+ klippy/extras/temperature_fan.py | 13 +++++++------
+ klippy/webhooks.py               | 15 ++++++++-------
+ 2 files changed, 15 insertions(+), 13 deletions(-)
+
+diff --git a/klippy/extras/temperature_fan.py b/klippy/extras/temperature_fan.py
+index 757b494..a358d2c 100644
+--- a/klippy/extras/temperature_fan.py
++++ b/klippy/extras/temperature_fan.py
+@@ -4,8 +4,6 @@
+ #
+ # This file may be distributed under the terms of the GNU GPLv3 license.
+
+-import numpy as np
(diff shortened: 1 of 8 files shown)


pdscomp:

- `-X no_debug_ranges`: "seems like a good change"
- `MALLOC_ARENA_MAX=1`: "sounds scary; could have side effects."
- Precompiled pyc: disagrees. "Don't see how it helps other than bootup times" / "wouldn't pyc files exist after first boot anyway" / "doesn't increase boot times by *that* much — more spots where stuff can go wrong." (Note: CC1 rootfs is read-only squashfs; pyc cannot be written on-device.)
- Same objections apply to `moonraker-init-d`.
- General numpy note: Python can be efficient if numpy data is handed between functions as pointers — suspects several "Python slow" paths are fixable without rewrites.

pdscomp:
Full application/runtime test artifact: https://gist.github.com/pdscomp/a062755da204c914c072de3391ea6c1d

It covers the build, flash/reboot validation, NumPy import behavior, real X/Y input-shaper runs, full calibration macro (PID and load-cell tare), the known bed-mesh drift limitation, memory observations, and rejected experiments.

pdscomp:

PR #259, opened 2026-07-19
calibration: skip saved stages, persist shaper results, stream accel writes

pdscomp:
# calibration: skip saved stages, persist shaper results, stream accel writes

Split from #252. **Draft — commits not yet merged into this branch.**

## Why this PR

Three real bugs/behaviors in the calibration path: (1) `SAVE_CONFIG` silently did not persist shaper results set by Rusty/Classic via `SET_INPUT_SHAPER`; (2) the boot/check flow re-ran every calibration stage even when saved results were current; (3) raw accelerometer CSV writes doubled peak memory by flattening batches into a full samples list first.

## Justification / measurements

**TBD** — isolated idle/calibration numbers with only this change set. (The +6.04 MiB `MemAvailable` idle improvement reported in #252 was measured on the full series and is not attributable to this PR alone.)

Verified behaviors:

- Root cause of lost shaper persistence: `input_shaper.py` `cmd_SET_INPUT_SHAPER` never called `configfile.set()`, so `SAVE_CONFIG` saw no dirty `[input_shaper]` section. Fixed.
- `CALIBRATE_ALL` defaults `FORCE=1` (full recalibration — old behavior preserved); `CHECK_CALIBRATION` passes `FORCE=0` to run only missing stages. Prompt text only asks operators to clear the chamber/bed when a stage actually requires it.
- adxl345 `write_to_file()` streams CSV directly from message batches; no duplicate full-capture list in RAM.

## Commits (to be cherry-picked)

- `7e5533f` calibration: skip stages with saved results
- `49d55d3` kalico: persist input-shaper settings and stream raw accelerometer writes

## Risk / rollback

Low. `FORCE=1` restores the historical always-full-recalibration behavior as default for the explicit macro; per-stage skips only affect the auto/check path. Rollback = revert macros + patches.


pdscomp:

- `0002-persist-set-input-shaper.patch`: doesn't understand what it fixes; questions why `SET_INPUT_SHAPER` is even needed.
- `0003-adxl345-stream-raw-writes.patch`: "I don't see this patch helping much — have you measured this?" (TBD number.)
- `calibration.cfg`: too much code to accomplish the goal; disagrees with merging full + partial paths; prefers user picks full-calibrate OR individual stages.
- Hesitant about `SAVE_CONFIG RESTART=0` between full-calibrate steps — wants users to check results before saving; auto-save only useful after crash recovery.
- Opposes auto-retry-on-crash: "if something crashes the why should be investigated, not bruteforced till it worked."
- Wants to hold off calibration changes — still merging lizard's proposals into that area.

pdscomp:

PR #260, opened 2026-07-19
Reset grumpyscreen config when needed

jamesturton:
Follows the logic from the klipper config reset, except it removed the config from the overlay rather than commenting the modifications out.
Maybe at some point we should also use `CONFIG_OVERRIDE_FILE`, but that seems a little bit overkill right now


suchmememanyskill (approved):

suchmememanyskill (approved):
Code looks overall fine, have not tested it personally

jamesturton:
Tested working on my machine

PR #261, opened 2026-07-20
cc1: harden RTL8821CU Wi-Fi across warm reboots

pdscomp:
## Problem

The Elegoo Centauri Carbon 1 RTL8821CU Wi-Fi adapter intermittently failed after warm reboot:

- Early failures occurred during USB enumeration, including `SET_ADDRESS` returning `-EPROTO`.
- Later captures showed the adapter enumerate and bind, then fail rtw88 vendor control transfers with `-EPROTO` during firmware startup.
- Once wedged, the old path issued more than 1,000 failing register controls and could remain broken across soft resets.

The original failure reproduced every 3-5 soft reboot/flash cycles on the CC1 original mainboard (not the updated wifi chip one).

### linux: add slow-reset quirk for the CC1 USB hub

Marks the CC1 QinHeng `1a86:8091` parent hub with `USB_QUIRK_HUB_SLOW_RESET`. USB core gives downstream devices an extra 100 ms after port reset, directly improving the pre-enumeration warm-reset failure without cycling the whole hub.

The initial hub-quirk plus shutdown-unbind experiment reached roughly 10 cycles before a different, post-bind `-EPROTO` failure was captured.

### rtw88: retry failed USB register controls

Retries only register-control `-EPROTO` failures after 5, 20, 80, 320, and 1280 ms. Successful transfers have no added delay. A terminal failure latches the control transport as broken, preserving the triggering operation/address and suppressing the previous thousand-message error storm.

This replaces a discarded blanket 5 ms delay experiment, which added about 18 seconds to Wi-Fi startup and still failed after 19 loops.

### rtw88: serialize USB vendor controls

Adds the vendor-request mutex used by Realtek's 8821cu driver. It keeps a register request and its RTL8821C `0x4e0` reg-sec companion atomic and serializes firmware-page controls on the same mutex, preventing control requests from interleaving. Firmware fragments are not individually retried because the full download is checksummed.

## Validation

- Clean rtw88 build: 1,143 tasks passed; both patches applied without fuzz.
- Full CC1 SWU build on current `main`: 6,521 tasks passed.
- Combined production fixes plus an experimental shutdown hook: 44 consecutive reboot loops.
- Final image with only the kernel/rtw88 changes and no shutdown hook: **540 consecutive successful reboot loops** on `carbon-devu`.
- No fixed successful-path delay remains.

The reboot counts validate the combined final patch set; the driver patches were not individually ablated.

### DIFF
--- a/meta-opencentauri/images/files/sw-description
+++ b/meta-opencentauri/images/files/sw-description
@@ -59,7 +59,7 @@ software =
                 },
                 {
                     name = "setargs_mainline";
-                    value = "setenv bootargs console=ttyS0,115200 earlycon=uart8250,mmio32,0x02500000 root=${mmc_root} rootwait consoleblank=0 usbcore.old_scheme_first=1";
+                    value = "setenv bootargs console=ttyS0,115200 earlycon=uart8250,mmio32,0x02500000 root=${mmc_root} rootwait consoleblank=0 usbcore.old_scheme_first=1 usbcore.use_both_schemes=0";
                 },
                 {
                     name = "bootcmd_";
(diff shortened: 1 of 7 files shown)


pdscomp:
Full Wi-Fi investigation log and test history:

https://gist.github.com/pdscomp/c207506c250913d83c703178d089b901

pdscomp:

PR #262, opened 2026-07-20
loadcell: prevent corrupt HX711 frames aborting probes

### LINKED ISSUE #241: "Error during homing probe: Load Cell Probe Error: force exceeded drift_safety_limit before triggering!"
I'm playing with the new nightly, and am running into issues with the new HX711 driver.

Occasionally, I am able to create a bed mesh, but roughly 80% of the time it fails at some random point measurement and displays the error message in the title. I tried taring, calibrating and then even tested the bed with different weights (1 and 3KG) and it reads the weights quite accurately.

I'm suspecting the error message above is tied to some dropped data between the bed level MCU and the main. I cannot correlate them exactly to the same timing, however the main log shows many warns for ignored packets:

```
[clocksync][244][Warn][59892559]:Ignoring clock sample sent_time[clocksync][244][Warn][59899749]:Ignoring clock sample sent_time:59899.747: freq=84002032.449615 diff=19322.299805 stddev=1421.186 strain_gauge_mcu
[clocksync][244][Warn][59921101]:Ignoring clock sample sent_time:59921.100: freq=83999734.163223 diff=11472.746094 stddev=1853.924 stm32
[clocksync][244][Warn][59943247]:Ignoring clock sample sent_time:59943.247: freq=199999999.997650 diff=28611.662109 stddev=2965.132 mcu
[clocksync][244][Warn][59951611]:Ignoring clock sample sent_time:59951.611: freq=83999733.772435 diff=12951.806641 stddev=1746.809 stm32
[clocksync][244][Warn][60026709]:Ignoring clock sample sent_time:60026.707: freq=84002031.239916 diff=11259.399414 stddev=950.749 strain_gauge_mcu
[clocksync][244][Warn][60032615]:Ignoring clock sample sent_time:60032.613: freq=84002031.083306 diff=9058.394531 stddev=929.871 strain_gauge_mcu
[clocksync][244][Warn][60033299]:Ignoring clock sample sent_time:60033.299: freq=83999733.582008 diff=32395.714844 stddev=1591.498 stm32
[clocksync][244][Warn][60045110]:Ignoring clock sample sent_time:60045.109: freq=83999733.900935 diff=8506.712891 stddev=1673.864 stm32
[clocksync][244][Warn][60047377]:Ignoring clock sample sent_time:60047.375: freq=84002031.197762 diff=18669.646484 stddev=882.314 strain_gauge_mcu
[clocksync][244][Warn][60056921]:Ignoring clock sample sent_time:60056.920: freq=83999734.797197 diff=10745.286133 stddev=1739.437 stm32
[clocksync][244][Warn][60110365]:Ignoring clock sample sent_time:60110.363: freq=84002032.165548 diff=12789.617188 stddev=1162.209 strain_gauge_mcu

```
In addition, klipp.log is producing these messages while just sitting idle (and also during leveling operations) and I believe this is the trigger  of the drift safety limit:

```
load_cell_probe: glitch dropped at t=2501.483; channel(s) [1] jumped (now=[-155235, -1, -62751, 474210] last=[-155168, -310564, -62757, 474271])
load_cell_probe: glitch dropped at t=2507.727; channel(s) [0] jumped (now=[-1, -310749, -62435, 474152] last=[-155181, -310757, -62686, 474145])
load_cell_probe: glitch dropped at t=2509.260; channel(s) [1] jumped (now=[-155137, -1, -62600, 474298] last=[-155185, -310779, -62539, 474366])
load_cell_probe: glitch dropped at t=2542.416; channel(s) [1] jumped (now=[-154917, -1, -62082, 474642] last=[-154707, -310348, -61933, 474616])
load_cell_probe: glitch dropped at t=2592.866; channel(s) [0] jumped (now=[-1, -310064, -61460, 475185] last=[-154207, -310151, -61475, 475140])
load_cell_probe: glitch dropped at t=2600.950; channel(s) [0] jumped (now=[-1, -309825, -61667, 475141] last=[-154023, -309835, -61705, 475248])
load_cell_probe: glitch dropped at t=2662.799; channel(s) [1] jumped (now=[-153409, -1, -60962, 476215] last=[-153481, -309103, -61052, 475995])
load_cell_probe: glitch dropped at t=2684.625; channel(s) [0] jumped (now=[-1, -309159, -60895, 476382] last=[-153252, -309040, -60912, 476291])
load_cell_probe: glitch dropped at t=2692.967; channel(s) [0] jumped (now=[-1, -309131, -60652, 476287] last=[-153239, -309116, -60595, 476300])
load_cell_probe: glitch dropped at t=2706.176; channel(s) [0] jumped (now=[-1, -308783, -60913, 476374] last=[-152967, -309000, -60683, 476265])
load_cell_probe: glitch dropped at t=2707.894; channel(s) [1] jumped (now=[-152854, -1, -60842, 476441] last=[-152855, -308916, -60758, 476501])
load_cell_probe: glitch dropped at t=2709.549; channel(s) [0] jumped (now=[-1, -308815, -60844, 476578] last=[-152915, -308702, -60821, 476556])
```

I am not sure why it is only channels 0 and 1. It doesn't make a lot of sense to me. Perhaps a glitch in the Bed MCU firmware?

My short term plan is to revert to the original [load_cell_probe] definitions from .7 to see if it solves the issue, but i thought I'd throw this up here as a heads up.

pdscomp:
## Problem

Intermittent `Load Cell Probe Error: force exceeded drift_safety_limit before triggering!` failures occur during probing and heated bed meshes. The same logs contain framing-valid HX711S samples where one channel becomes exactly `-1`. The host-side detector identifies these jumps, but the MCU made the probe endstop decision first.

The MCU compared only the summed sample against a hard-coded 200,000-count threshold, while this printer's configured 1,000 g drift guard is 105,000 counts. A channel jump between those values could pass the MCU filter and falsely trip the safety guard. This explains the machine/channel-dependent behavior in #241.

## Fix

- Track the last good count for each HX711S channel on the bed MCU.
- Quarantine one suspicious per-channel jump using the existing 100,000-count physical-jump bound.
- Accept a return-to-baseline sample as an isolated corrupt frame.
- Accept two consecutive samples near a new value as a genuine sustained force change, so real contact cannot be filtered indefinitely.
- Keep raw samples in the bulk buffer for host diagnostics and leave drift/force safety limits unchanged.
- Bump the Kalico package revision from `r7` to `r8` so the boot-time flasher installs the changed bed firmware.

## Validation

- STM32F401 bed firmware compiled and linked successfully.
- Full Yocto `opencentauri-upgrade` image: all 6,521 tasks succeeded.
- `SAVE_CONFIG` was issued before flashing; the four-channel tare survived both image reboots.
- Cold `G28` completed despite a naturally occurring channel-2 `-1` frame.
- `LOAD_CELL_DIAGNOSTIC` completed.
- Two complete 9x9 `BED_MESH_CALIBRATE PROFILE=default` runs completed and saved successfully. Both runs encountered additional channel-1/channel-2 `-1` frames, but neither produced a new drift-safety abort, MCU shutdown, or communication error.

Detailed investigation, the original #241 sample, arithmetic, implementation rationale, and raw test observations are in [LOADCELL.md](https://github.com/OpenCentauri/cosmos/blob/doom/loadcell/LOADCELL.md).

Closes #241


pdscomp:
Verification: 20 successful full-bed probe completions on carbon2u.

Breakdown:
- 10/20 would have aborted on the old code path
- 10/20 were clean on the old code path

Artifact: https://gist.github.com/pdscomp/1ac18603d288a48fc9c8e50bf79e65e1

Details are in LOADCELL.md in the gist; source log was /board-resource/klippy.log.

pdscomp:
Verification update for the HX711S glitch-filter follow-up:

- Restored first-sample forward in the `!have_last_counts` seed path so the first valid sample after query still reaches `load_cell_probe`.
- This preserves the old behavior for future mid-probe restart cases instead of silently dropping the first sample.

Regression test on carbon2u:
- Built a full cc1 eMMC image.
- Flashed the image to carbon2u.
- Forced the bed firmware refresh via `/etc/init.d/klipper-firmware-bed` after clearing `/etc/bed.ver`.
- Rebooted and ran 10/10 full-bed mesh probes successfully.

The log now shows 10 `Mesh Bed Leveling Complete` entries after the flash cycle.

pdscomp:
### Additional probe failure observed on carbon2u

During repeated calibration testing on `PR-262-b5d`, with the chamber at approximately **60 C** and the bed at **100 C**, we repeatedly encountered a separate failure mode from the original `drift_safety_limit` false abort:

```text
Probe samples exceed samples_tolerance
```

At one probe location, repeated readings differed by about 0.0557 mm (`-1.451570` vs `-1.507287`), just beyond the configured 0.05 mm tolerance. The run also logged a quarantined channel-2 `-1` glitch frame. This is documented in [issue #241](https://github.com/OpenCentauri/cosmos/issues/241#issuecomment-5035543016) and should be investigated as a possible temperature-dependent load-cell/mechanical repeatability or signal-integrity issue. No config was saved from the failed run.

jamesturton (approved):
I've looked through the changes and I think it seems like a good approach. Given both you and @lpaln have tested and seen markable improvements I think this should be merged. I have applied the patch you made to the `hx711s-new` branch of our kalico fork and updated the srcrev to match.

bniggemyer:
ran several prints under #262 build -- all clean!

jamesturton:
> ran several prints under #262 build -- all clean!

Great! Thanks for the feedback!

pdscomp:
Follow-up testing found additional failure cases on carbon2u in the latest `PR-262-b5d` build, especially during high-temperature probing (user-reported chamber temperature >50 °C with a 105 °C bed).

Two bed-probe attempts reported `Probe samples exceed samples_tolerance`. A later long-running attempt was followed by Klippy losing the load-cell bulk-sensor response (`Unable to obtain 'sensor_bulk_status' response`) and reconnecting. The retained log contains 192 dropped invalid/discontinuous HX711 channel frames, including the recurring `-1` channel values.

This is different from the original `drift_safety_limit` failure addressed by this PR. The per-channel classifier prevented the previously demonstrated false drift-safety abort during earlier cold/hot mesh validation, but the underlying invalid ADC frames remain possible and the latest hot run exposed probe-tolerance and sensor-stream failure modes. The onboard chamber sensor read 31.8 °C at the captured MCU failure, so the reported >50 °C chamber condition needs controlled confirmation with an independent measurement near the load-cell board.

I documented the new evidence and follow-up plan in [issue #241](https://github.com/OpenCentauri/cosmos/issues/241#issuecomment-5047841786) and updated the [load-cell investigation gist](https://gist.github.com/pdscomp/1ac18603d288a48fc9c8e50bf79e65e1). We should investigate thermal/acquisition robustness without weakening the existing safety limits.

lpaln:
@pdscomp 

I am going to be running some additional tests over here as well. Since the -1 values are only appearing on the long wired sensors in my small sample, I have ordered some nickel-zinc ferrite rings to attempt to suppress any high frequency noise in the long sensor wires. While this is basically darts at a wall at this point, My prime suspicion is that the cable length is close to a multiple of the wireless wavelength (seems to be in range of 5x) and that the cables are picking up interference from the built in Wifi trasmitter. Per AI, the nickel-zinc composition rings are specifically tuned to filter out noise in the wireless frequency spectrum. I'll drop my findings up here once I have any.

Issue #150, opened 2026-04-17
Calibration issues


Ididnttouchthat:
<img width="4000" height="3000" alt="Image" src="https://github.com/user-attachments/assets/9ebad040-0af6-404d-bf19-6324481976e3" />
<img width="4000" height="3000" alt="Image" src="https://github.com/user-attachments/assets/5d7720bd-994e-49e9-b8c3-480631ff4662" />
<img width="4000" height="3000" alt="Image" src="https://github.com/user-attachments/assets/ae95c3b4-8ea3-48b6-a343-70932413fef8" />
<img width="4000" height="3000" alt="Image" src="https://github.com/user-attachments/assets/527a7870-d59a-4f17-9bb5-8c8592e9f3d0" />

robertsamples:
Some text explanation would be helpful here, we don't know how these photos are related and what  you did previously in terms of calibration, or what version of cosmos you are using.

You should run the LOAD_CELL_TARE command with nothing on the bed  which is likely your issue with probing.

you should turn the fans off prior to running IS

Ididnttouchthat:
So once it initially booted it prompted for a full calibration started that, that when I got the error lis2dw It cooled down to 49c the hot end fan turned off but it never ran the input shapping, power cycled dismissed calibration ran bedmed mesh manually that's when I got the load cell probe error quick Google search adjusted the tare/baud rate based on Google went positive first still failed that's when I made it negative then it was actually able to run the bedmesh manual input shapper command finished no problem set z offset have not ran a print yet tho I kinda figured picture would be self explanatory that's why I didn'twrite any thing and this is my first time ever posting on GitHub but I figured I want this project to actually work so I'd throw my 2 cents in also using 0.0.4 the current one

robertsamples:
ok I would not adjust the baud rate as that will prevent the boards from communicating properly.  But if you were able to get bed mesh calibrated and z offset you should run save_config and then reboot the printer. Sometimes things can be a bit strange on the first boot or two after install. After that you can try to run IS again and it should work.

jbarron16:
I’m having a similar issue on the 0.0.6 update.

I originally started on 0.0.3 and was able to run the full calibration macro without any issues. After updating, the macro started failing during the Input Shaper section with this error:

WARNING: Active fans detected: extruder
Fan vibrations may affect input shaper calibration accuracy.
For best results, stop fans or wait for them to stop automatically.

Invalid lis2dw id (got ff vs 44).
[...]

PR #270, opened 2026-07-26
Klipper: Read from accelerometer to wake it before shaper calibration

jamesturton:
I think I was too hasty closing #150 as discord user spaget0119 has also been having this problem
https://discord.com/channels/1367538416539013122/1384643865885151322/1528494762674360320

I don't have the ability to test this right now, but I can't see why it would harm anything, and maybe it would help?


suchmememanyskill (approved):
Fine with just merging this and seeing if the issue stops for people

PR #272, opened 2026-07-27
Klipper: Update hx711s with new architecture

jamesturton:
Hopefully this fixes the `-1` readigs we were seeing in the previous driver now that we are reading the chips independently. This fix allows us to remove the `LOAD_CELL_CALIBRATE TARE=TRUE` hacks I added to `macros.cfg` as all probing moves call
`LoadCellPrimitives.tare()` unconditionally.
Also updating `sdo_pins` to `dout_pins` to be consistent with the datasheet and the hx71x driver.

Link to kalico branch: https://github.com/OpenCentauri/kalico/commits/rpmsg-with-new-hx71x/


jamesturton:
Thank you @pdscomp @jaimbo @Atomique13 and @harrym for testing!

Issue #59, opened 2026-04-02
PID tuning hotend autocycles repeated/times out with hotend fan off


robertsamples:
**Firmware**: COSMOS 0.0.1

**Hardware modifications**: Constellation extruder with TZ3.0 hotend, melt zone extender, V6 nozzle; Gamma cowling

**Symptom**: during PID tuning hotend will repeat heating cycles a large number of times, in one case 20 and another 30 before timing out. This only occurs in relatively low heat loss conditions and I did not observe it with the model fan on. I am unsure if it would be observed with the stock hotend and cowling- airflow from gamma cowling does not impinge on nozzle as much as stock.

**Reproduction steps**: PID tune hotend with model fan off, preheated heatbed bay be required, may not occur with stock hotend/cowling

<img width="905" height="742" alt="Image" src="https://github.com/user-attachments/assets/734e026d-6030-430b-9eb7-ec0b27d312f1" />

robertsamples:
@jaimbo fascinating by stock is this pulled from the stock firmware or do you mean the stock hotend? was the part cooling fan on at all?

robertsamples:
@jaimbo 

weird. I'm not sure whats going on with mine. maybe I'll try repasting the thermistor and heater

robertsamples:
@Atomique13 @jamesturton 

I believe I have a fix for this https://github.com/OpenCentauri/cosmos/pull/282

Atomique13:
Fixed in #

PR #282, opened 2026-08-11
Fix issue #59- bump thermistor smooth time

robertsamples:
Issue 59 (https://github.com/OpenCentauri/cosmos/issues/59) appears to be caused by higher powered/lower thermal mass hotends causing issues with PID tuning with the default smooth time of 1.0 seconds. I tested and found that increasing this value to 2.0 seconds provided accurate results and eliminated noise that caused PID tuning to fail. On my testing tuning failed after 30 cycles with 1.0 seconds and succeded reliably after the expected 6 cycles with 2.0 seconds.


Bad-Hackerman:
image can confirm. higher temperatures seem to help smoothing it out, but at low temperatures it's all over the place

Bad-Hackerman:
image
same settings just with a smooth time of 2s instead of 1s

Bad-Hackerman (approved):

PR #297, opened 2026-08-22
fluidd/mainsail: default off built-in Screws Tilt Adjust dialogs

pdscomp:
## TL;DR (plain language)

When you run screw-based bed leveling, Fluidd and Mainsail each pop **their own** built-in "Screws Tilt Adjust" window on top of the new Klipper-prompt leveling flow from #248. Two competing UIs showing the same numbers is garbage soup. This PR turns the frontends' popups **off by default** in the shipped image, so the prompt-based flow is the only thing on screen. Anyone who misses the old window can re-enable it (Fluidd: Settings → Toolhead; Mainsail: Settings → Interface).

**Pairs with #248** (@Bad-Hackerman's macro side—this PR is the frontend counterpart). Ship together: without #248, stock `SCREWS_TILT_CALCULATE` falls back to console-text output only.

**Lingo:** STC = `SCREWS_TILT_CALCULATE` (the Klipper command that probes over each bed screw). "The dialog" = each frontend's native Screws Tilt Adjust popup, which auto-opens whenever STC results arrive.

## Files changed

| File | Change |
|---|---|
| `meta-opencentauri/recipes-apps/fluidd/files/opencentauri-fluidd-theme/apply-opencentauri-fluidd-theme.py` | New `patch_screws_tilt_default()` flips Fluidd's shipped store default off. New `verify_final()` independently re-checks the end state of **every** patch (theme included). Also removed a now-dead silent-success path in `patch_state_chunk`. |
| `meta-opencentauri/recipes-apps/mainsail/files/apply-opencentauri-mainsail-patches.py` | **New**—flips Mainsail's shipped store default off, plus the two `?? true` getter fallbacks (covers settings DBs that predate the key), with exact-occurrence assertions and an end-state verify pass. |
| `meta-opencentauri/recipes-apps/mainsail/mainsail_2.18.2.bb` | Runs the patch script in `do_install` (adds `inherit python3native` + the script to `SRC_URI`). Mirrors the fluidd recipe's existing pattern. |

## Why patch minified bundles, and why it's safe (check methodology)

Fair question—string surgery on minified JS is normally where maintainability goes to die. Here's why this instance is the right tool, and why it can't silently rot:

1. **There is no config-file route (verified against the shipped zips).** Fluidd's `config.json` carries only `blacklist`/`endpoints`/`hosted`/`themePresets`; Mainsail's carries only `defaultLocale`/`defaultMode`/`defaultTheme`/`hostname`/`instances*`. Neither feeds `uiSettings` defaults. The defaults exist *only* in the minified store state inside the release zips, so package-install time is the one place we control them.
2. **The match anchors are API-stable by construction.** We match on the persisted setting keys (`showScrewsTiltAdjustDialogAutomatically`, `boolScrewsTiltAdjustDialog`). Those strings are serialized into users' moonraker settings DBs—upstream can't rename them without orphaning every existing user's saved preference. Minification mangles everything *around* the keys; the keys themselves must survive. We never reference chunk content-hashes or minified variable names (globs only).
3. **Exact-occurrence assertions.** Each replacement asserts its expected count (1 for both defaults, 2 for Mainsail's fallbacks). A newer upstream build that grows a second defaults object—or drops one—fails `do_install` instead of patching the wrong occurrence.
4. **Independent end-state verify.** After patching, each script re-reads the final webroot and asserts the desired state (off-forms present exactly N times, on-forms absent). Fluidd's verify also re-checks every theme marker, closing a silent hole in the existing theme patcher.
5. **It runs inside `do_install`, so it survives updates by construction.** The zip is sha256-pinned; upstream can't drift under us. A version bump that moves the strings turns the bump PR's build red with the offending path in the error—loud failure aimed at exactly the person who can fix it, instead of a quietly unpatched dialog shipping to printers.
6. **Doubles as the version-bump smoke test.** `unzip mainsail.zip && python3 apply-opencentauri-mainsail-patches.py webroot/`—a non-zero exit means the patch needs porting. No separate tooling to maintain.

## Verification

- `bitbake -c cleansstate fluidd mainsail && bitbake fluidd mainsail`—all 1903 tasks green.
- Packaged webroots: fluidd `state-*.js` has exactly 1× `showScrewsTiltAdjustDialogAutomatically:!1` and 0× `:!0`; mainsail `index-*.js` has 1× `boolScrewsTiltAdjustDialog:!1`, 2× `==null?!1:e` fallbacks, 0 remaining on-forms; OpenCentauri theme markers intact.
- Fail-hard proven: re-running either script against an already-patched (or string-drifted) tree raises `RuntimeError` with the path.
- Reviewed (twice) with a severity-ranked code review pass; findings (`encoding="utf-8"` hygiene, one dead code path) fixed and re-verified.

## Caveats

- Printers whose moonraker DB already holds an explicit `true` for these keys keep the popup—saved user settings correctly win over shipped defaults. Fresh flashes, and anyone who never touched the toggle, get the new default.
- Re-enabling is one settings toggle in either frontend, no reflash needed.

### DIFF
--- a/meta-opencentauri/recipes-apps/fluidd/files/opencentauri-fluidd-theme/apply-opencentauri-fluidd-theme.py
+++ b/meta-opencentauri/recipes-apps/fluidd/files/opencentauri-fluidd-theme/apply-opencentauri-fluidd-theme.py
@@ -174,8 +174,6 @@ def patch_main_bundle(webroot):
 def patch_state_chunk(webroot):
     """Patch the initial Vuex state so OpenCentauri is the default theme."""
     chunks = sorted((webroot / "assets").glob("state-*.js"))
-    if not chunks:
-        return  # Not all builds have this chunk
 
     old_theme = "theme:{isDark:!0,logo:{src:`logo_fluidd.svg`},color:`#2196F3`,backgroundLogo:!0}"
     new_theme = f"theme:{{isDark:!0,logo:{{src:`{LOGO_SRC}`}},color:`{PRIMARY_COLOR}`,backgroundLogo:!0}}"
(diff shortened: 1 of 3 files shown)


pdscomp:
@Bad-Hackerman you are the one who requested this, assigning to you to do what you will with it

PR #313, opened 2026-09-05
Run COSMOS updates through Moonraker's update_manager

kyleinoregon:
Replaces #308, following the discussion there: consistency across UIs by using the update_manager API that Mainsail and Fluidd already implement. The screen side is pellcorp/grumpyscreen#289.

Reworked after the first review: everything about updates now lives in Moonraker. No shell helper, no Klipper macros, no startup check script, and the component no longer touches update_manager's table beyond adding its own entry.

## What changes

- **`cosmos_update.py`**, a Moonraker component placed into the Moonraker source tree by the recipe (`subdir=`), so the installed tree is the same as the source. It registers a `cosmos` entry with `update_manager`: installed version from `/etc/issue`, channel from `config-manager`, latest version from the GitHub releases list on the stable channel or commits behind `main` on nightly. The update step runs `update-cosmos` and streams its output line by line to the clients.
- **`0001-update_manager-skip-placeholder-applications.patch`**: update_manager always creates `klipper` and `moonraker` entries, and on this image neither is a git checkout, so they are bare placeholders with an empty status that Mainsail and Fluidd render as empty rows. The patch leaves entries without a status out of the status response and the refreshed notification (two hunks). Worth sending upstream as well.
- **`moonraker-readonly.conf`** enables `[update_manager]` (system updates off, refresh once a day) and the `[cosmos_update]` section.
- **`update-cosmos`** now prints what it is doing: download progress in 20 % steps, install, reboot, and a plain failure line if the download or the flash fails. The curl and swupdate chatter goes to `/board-resource/update-cosmos.log` instead of the user's screen. URLs, `flash`, cleanup, reboot and exit codes are unchanged.
- **Removed**: the `check-update` script and recipe, the `CHECK_FOR_UPDATES` / `UPDATE_COSMOS` shell commands, the `_UPDATE_COSMOS` macro, the startup `check_for_updates_startup` delayed gcode, the "Update Available" special case in `_SHOW_PROMPT`, and the `check_for_updates` option in `cosmos.conf`. update_manager's own refresh replaces the check, and the web UIs show the result in their update panel. A leftover `check_for_updates` line in a user's `cosmos.conf` is ignored by config-manager.

## What the user sees

- Mainsail and Fluidd: a "cosmos" row in the update panel with `26.08.0 > 26.09.0` and a working Update button when a release is out, a release-notes link, and the standard update dialog with the live log during the update. The dialog appears whichever UI or button started the update.
- Screen: with pellcorp/grumpyscreen#289, the same messages in a dialog on the panel, ending on "COSMOS update installed, the printer is rebooting". The startup "Update Available" prompt is gone; if a screen-side notice is wanted, it can be added on the grumpyscreen side from `notify_update_refreshed`.

## Tested

On a Centauri Carbon running 26.08.0:

- Component loaded with Moonraker 0.10 as shipped; status reports `26.08.0` installed / `26.08.0` latest from GitHub; with the patch the placeholder klipper/moonraker rows are gone.
- Dry-run update through `machine.update.client`: every line streamed to a websocket client and to Mainsail's update dialog, completion flag at the end.
- `update-cosmos` run against the real release asset with `flash` and `reboot` stubbed: download progress lines at 20 % steps, install and reboot lines, chatter in the log file, swu file cleaned up.
- A real update through the whole path, started from the screen's Update button with the #289 client: 26.08.0 re-flashed to the other slot (boot slot flipped A -> B, swupdate's post-update steps in the log), progress shown on the screen and in Mainsail, automatic reboot, printer back in about a minute with `printer.cfg` byte-identical and all calibration loaded.
- The final line carries update_manager's completion flag, so clients know when to close their dialog.

The recipe's grumpyscreen SRCREV is not bumped here; pellcorp merges #289 once this is in, and the bump can follow.

The docs page that mentions `check_for_updates` (troubleshooting/questions.md in OpenCentauri/OpenCentauri) needs a follow-up once this lands.


meta-opencentauri/recipes-apps/moonraker/moonraker_0.10.0.bb
```diff
@@ -13,6 +13,7 @@ SRC_URI = " \
     file://moonraker-init-d \
     file://moonraker.conf \
     file://moonraker-readonly.conf \
+    file://cosmos_update.py \
```
jamesturton: Use subdir, we want the source to be coherent

meta-opencentauri/recipes-apps/moonraker/moonraker_0.10.0.bb
```diff
@@ -70,6 +71,8 @@ do_install() {
     # Install moonraker python package
     install -d ${D}${datadir}/moonraker
     cp -r ${S}/moonraker ${D}${datadir}/moonraker/
+    # COSMOS updater for update_manager
+    install -m 0644 ${WORKDIR}/cosmos_update.py ${D}${datadir}/moonraker/moonraker/components/
```
jamesturton: Remove this, see comment above

meta-opencentauri/recipes-apps/grumyscreen/files/grumpyscreen.cfg
```diff
@@ -44,10 +44,14 @@ port: 80
...
 
+# The update button drives the cosmos entry of Moonraker's update_manager
+[update_manager]
+application: cosmos
+
 [commands]
 factory_reset_cmd: /usr/bin/factory-reset
 gui_restart_cmd: /etc/init.d/gui-switcher restart
-cosmos_update_cmd: /usr/bin/update-cosmos
+cosmos_update_cmd: /usr/bin/cosmos-update-start
```
jamesturton: Why are you splitting it out over grumpyscreen logic and shell scripts?
The whole point of this was for grumpyscreen to use the existing moonraker API rather than including lots of hacks for cosmos

meta-opencentauri/recipes-apps/klipper/files/shell.cfg
```diff
@@ -93,8 +93,8 @@ timeout: 5
 verbose: False
 
 [gcode_shell_command UPDATE_COSMOS]
```
jamesturton: I would like all these update macros to be removed and for everything regarding updates to be handled in moonraker, not klipper

meta-opencentauri/recipes-data/update-scripts/files/cosmos-update-start
```diff
@@ -0,0 +1,6 @@
+#!/bin/sh
+# Start a COSMOS update through Moonraker's update_manager and return at once.
+# Every UI that speaks the update_manager API then shows the progress; the
+# update itself runs under Moonraker, outside Klipper's command queue.
+curl -s -X POST "http://localhost/machine/update/client?name=cosmos" >/dev/null 2>&1 &
```
jamesturton: I do not like this at all. Grumpyscreen should use the existing moonraker connection. Splitting it out into a shell script makes no sense here

jamesturton (changes requested):
I've just done a first pass at reviewing this, but there a quite a few major issues so I haven't gone into too much detail.

PR #325, opened 2026-09-12
Bump kalico version

jamesturton:
Needs testing before merging.
Specifically that this version includes this commit https://github.com/OpenCentauri/kalico/commit/6b1822bc76c54faaa1b95331d886a8b951330882 which is supposed to reduce latent memory requirement so a comparison of memory usage between nightly and this PR would be good.


Atomique13 (approved):
works on my printer and even done 1 ams print. Please check discord for more data. love ya <3

jamesturton (commented):
@Atomique13 has tested this, posting his result here. Ram reduction of about 3mb

Before:


After:

PR #329, opened 2026-09-17
Wipe after print; reconcile KICK and CLEAN_NOZZLE; Don't unload after cancel

suchmememanyskill:


suchmememanyskill (commented):
TODO: There's now 2 purge macros in macros.cfg, that's not good

dylanparry:
The new purge macro doesn't appear to be turning the part cooling fan off when actively extruding. Won't that result in long strings of cooled filament rather than curled up blobs?

suchmememanyskill:
Good catch but in practise it won't matter much as the purge chute of the cc is too large/low. Will fix though

suchmememanyskill:
Lizard approves

PR #333, opened 2026-09-19
Make AFC fallback to default material temp when orca mapped material temp doesn't exist

suchmememanyskill:


Atomique13 (approved):
I have no idea what I am doing :dogmeme:

PR #334, opened 2026-09-19
images: add CC2 eMMC wic layout with matching GPT partitions (split cc1/cc2 eMMC wic file), update cc2 fstab

pdscomp:
PR to update the partitioning scheme to match CC2 eMMC and update fstab for mounting correct partitions.

1. Split opencentauri-mmc-image.wks.in to files/wic/opencentauri-cc1-mmc-image.wks.in and files/wic/opencentauri-cc2-mmc-image.wks.in
2. Update opencentauri-image-mmc.bb with required updates
3. Update the fstab (fstab.elegooo-centauri-carbon2) to mount UDISK to /opt/usr (like CC2) with bind mount to /user-resource for compatibility with existing Cosmos configs/settings
4. Correctly mount boot-resource partition as well
5. board-resource doesn't exist, it just gets ignored for cc2

Still to do: If we want to be able to seamlessly flash between cosmos and Elegoo updates, will need to manipulate and sign the swu in some fun ways (spoof version #, add signature in weird cc2 way, etc.) Will be different PR!

Testing: Successfully built both cc1 and cc2 eMMC wic/images, validated against partition layouts.


jamesturton:
I have pulled this into my branch! Thank you! :rocket:

